PR #21, opened 2016-02-27
Asset loader

jaredpalmer:
This adds cache busting (via hashes) for js files in production as well as some style changes to demonstrate Aphrodite.


jaredpalmer:
LGTM

PR #26, opened 2016-03-01
should fix the 404 in production

rowellx68:
When running in production, a 404 is thrown because of incorrect paths.


brunowego:
:+1:

PR #28, opened 2016-03-01
Fix item links on the about page

sepal:
The links on the about page are broken, because the href uses `{item.link}` as string. This pull request fixes this issue.


jaredpalmer:
LGTM

Issue #27, opened 2016-03-01
Unexpected key "currentPost"


Arueilen:
Hi, I found your boilerplate yesterday and it seems very cool !
But I have a bug when I want to run your example.
When I go to http://0.0.0.0:5000/ all works fine.
But if I go to http://0.0.0.0:5000/post/hows-business-boomin and then refresh the page, I have some errors into my Chrome and Firefox inspector: 
- Unexpected key "currentPost" found in initialState argument passed to createStore. Expected to find one of the known reducer keys instead: "posts". Unexpected keys will be ignored.
- Warning: React attempted to reuse markup in a container but the checksum was invalid. This generally means that you are using server rendering and the markup generated on the server was not what the client was expecting. React injected new markup to compensate which works but you have lost many of the benefits of server rendering. Instead, figure out why the markup being generated is different on the client or server:
 (client) <noscript data-reacti
 (server) <div class="root_fpza

I think it's a bug relate to asynchronous reducers. Why do you want asynchronous reducers ?

LASkuma:
Asyn reducers are used for code splitting.

jaredpalmer:
Fixed.

PR #30, opened 2016-03-03
Fix warning Warning: React attempted to reuse markup in a container b…

LASkuma:
This could fix the warning of 

`Warning: React attempted to reuse markup in a container but the checksum was invalid. This generally means that you are using server rendering and the markup generated on the server was not what the client was expecting. React injected new markup to compensate which works but you have lost many of the benefits of server rendering. Instead, figure out why the markup being generated is different on the client or server`

when directly visit post urls like [http://localhost:5000/post/the-pathway-to-more-success](url)

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -51,12 +51,12 @@ browserHistory.listen(location => {
 
     // Fetch deferred, client-only data dependencies:
     trigger('defer', components, locals);
+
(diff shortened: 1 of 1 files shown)


LASkuma:
This can fix the second problem mentioned in #27

LASkuma:
Would open a new PR.

PR #46, opened 2016-05-06
Move some dependencies to devDependencies and remove node-fetch.

mgmcdermott:
Some of the dependencies listed are only needed in development. I've also removed node-fetch as it did not seem to be used anywhere.


jaredpalmer:
Good catch on node-fetch. However I kept everything in dependencies, so that it's easy to deploy to Heroku and use GitHub sync.

PR #48, opened 2016-05-14
use axios instead of custom http-client

jaredpalmer:


gihrig:
Works for me 💯  

All tests pass, all features work as expected.

jaredpalmer:
Only locally, things get weird with CORS if you deploy to a real domain name

jaredpalmer:
@rowellx68 does this work for you?

rowellx68:
Only able to test locally and I still managed to encounter the CORS issue. `http://0.0.0.0:5000` vs `http://localhost:5000`.

On deployment, would it not be solved by providing the `WEBSITE_HOSTNAME` as you've set [here](https://github.com/jaredpalmer/react-production-starter/pull/48/commits/541e378db90346bfaf2e9af43debcb7bb0112dd1#diff-9eede41beb55de0e03a0aef05e8e95b9R6)?

jaredpalmer:
I will close and open this to request a new deploy

jaredpalmer:
It works on the server, but breaks because the client can't access the `WEBSITE_HOSTNAME`.

src/utils/HttpClient.js
```diff
@@ -1,66 +0,0 @@
-import request from 'superagent';
-import { canUseDOM } from 'fbjs/lib/ExecutionEnvironment';
-
-function getUrl(path) {
```
jaredpalmer: I think this can be fixed by moving this logic to an `axios.interceptor` which looks something like: 

```javascript
// Add a request interceptor
axios.interceptors.request.use(function (config) {
    // Do something before request is sent
    return config;
  }, function (error) {
    // Do something with request error
    return Promise.reject(error);
  });
```

jaredpalmer: So here's the rundown:

create axios clients on both client and server, setup the interceptors, then pass each instances to the `configureStore` , and finally create promise middleware that will accept the different clients...and we're done.

jaredpalmer:
This works now @rowellx68 @gihrig. I believe this will solve the cookie problem too, or at least make it trivial.

jaredpalmer:
carrying this to #63

PR #53, opened 2016-06-03
image and font support

jaredpalmer:
This PR adds support for including images and fonts in the application via `url-loader`. It also demos how to require an image inside of a react component.


jaredpalmer:
@rowellx68 @b2whats Mocha w/es6 doesn't like this implementation. Any thoughts?

jaredpalmer:
There is also some weirdness during the Image's load event.

jaredpalmer:
Meanwhile, it pretty much works..... https://spseo-pr-53.herokuapp.com/

justingreenberg:
> There is also some weirdness during the Image's load event.

it looks like server rendered markup is pointing to the wrong url...
```html
<img src="692adb35708d9279ca7eb56c71bfc102.png" class="logo_hn2i6o" alt="" data-reactid="5"/>
```
404: https://spseo-pr-53.herokuapp.com/692adb35708d9279ca7eb56c71bfc102.png

which is why broken link img flashes, however once mounted react renders the correct url:
```html
<img src="/build/static/692adb35708d9279ca7eb56c71bfc102.png" class="logo_hn2i6o" alt="" data-reactid="5">
```
200: https://spseo-pr-53.herokuapp.com/build/static/692adb35708d9279ca7eb56c71bfc102.png

any ideas?

b2whats:
hm, it's a strange approach for server-side rendering of the images

jaredpalmer:
@justingreenberg means that the server's webpack config probably needs adjustment. The other weird thing with all of this is that it generates two images.

justingreenberg:
adding public path to server webpack fixes issue, but two files are still being created....

```
    publicPath: '/build/static/'
```

b2whats:
in development mode we need use only file-loader for that would give the compiled file name in the html markup

file loader have one param:
```js
require("file?name=[hash].[ext]!./filename.js");
```

and we use node-hook in order to override the import function

```js
// simple example
import fs from 'fs'
import { interpolateName } from 'loader-utils'
const webpackConfig = require('......./webpack.config.dev').default

require.extensions['jpg'] = function (require, filename) {
  const content = fs.readFileSync(filename)
  const webpackName = interpolateName({resourcePath: filename}, '[hash].[ext]', {content}); // !!! we need in webpack config default hash param
  require.exports = webpackConfig.output.publicPath + webpackName
}
```

b2whats:
i am implement your example and as expected the server rendering is not working correctly in dev mode

```js
warning.js:45 Warning: React attempted to reuse markup in a container but the checksum was invalid. This generally means that you are using server rendering and the markup generated on the server was not what the client was expecting. React injected new markup to compensate which works but you have lost many of the benefits of server rendering. Instead, figure out why the markup being generated is different on the client or server:
 (client) x0xyai.1"><img src="/build/static/692adb
 (server) x0xyai.1"><img src="[object Object]" alt
```

also url-loader use file-loader, he use if file size > config value size

b2whats:
my attempt to make an automatic connection block hooks, but this draft
https://gist.github.com/b2whats/9f567b3b66859bd652ba920ab9a3a74f

PR #56, opened 2016-06-04
add React-hot-loader-3

b2whats:


src/client.js
```diff
@@ -34,9 +35,11 @@ let render = () => {
     // We need to have a random in development because of `match`'s dependency on
     // `routes.` Normally, we would want just one file from which we require `routes` from.
     ReactDOM.render(
-      <Provider store={store}>
+      <AppContainer>
+        <Provider store={store}>
           <Router routes={routes} history={browserHistory} key={Math.random()}/>
```
justingreenberg: random key is no longer necessary :)

package.json
```diff
@@ -59,6 +59,7 @@
     "react": "^15.0.2",
     "react-dom": "^15.0.2",
     "react-helmet": "^3.1.0",
+    "react-hot-loader": "^3.0.0-beta.1",
```
justingreenberg: we also must update to react-router v3-alpha to use this

jaredpalmer:
Closing this as state is 100% managed by Redux, I don't think we need RHL. Vanilla HMR will work fine.

PR #57, opened 2016-06-04
node-hooks solution

justingreenberg:
summary...
- added [node-hooks for resolving files in node](https://github.com/justingreenberg/react-production-starter/blob/351d83a596d70218a2ecda4615393fc9ba784d69/src/server/node-hooks.js)
- updated file/url loaders to use filenames and content hash for caching
- fixed some inconsistent styling, ie using es5 in webpack entry files for older runtimes 
- update configs to use loader query objects instead of query string syntax
- added simple `getPath` helper to webpack configs for relative path resolution
- moved react optimizations from root `.babelrc` into prod webpack config... maybe consider removing babelrc altogether?
- remove unnecessary `/build` path from public URI, [just use `/static` instead](https://github.com/jaredpalmer/react-production-starter/compare/images...justingreenberg:images?expand=1#diff-f286c15f5cb962ed226f2aae653ca1cfR88)
- update script tag for [main.js with `async` attribute](https://github.com/jaredpalmer/react-production-starter/compare/images...justingreenberg:images?expand=1#diff-f286c15f5cb962ed226f2aae653ca1cfR305) since vendor has entry logic

i wasn't able to workaround the duplicate file output to `build/server`, imo not a huge problem but something that we can create a ticket for and i will investigate further

let me know if you have any questions :)

edit: tests passed by importing node-hooks into server test file, which is a good sign

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,4 @@
 {
-  "presets": ["es2015", "react", "stage-0"]
+  "presets": ["es2015", "react", "stage-0"],
+  "plugins": ["transform-runtime"]
 }
--- a/.babelrc
+++ b/.babelrc
@@ -1,4 +1,3 @@
 {
-  "presets": ["es2015", "react", "stage-0"],
-  "plugins": ["transform-react-constant-elements", "transform-react-inline-elements"]
+  "presets": ["es2015", "react", "stage-0"]
 }
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
     "babel-loader": "^6.2.2",
     "babel-plugin-transform-react-constant-elements": "^6.5.0",
     "babel-plugin-transform-react-inline-elements": "^6.5.0",
+    "babel-plugin-transform-runtime": "^6.9.0",
     "babel-polyfill": "^6.5.0",
     "babel-preset-es2015": "^6.5.0",
     "babel-preset-react": "^6.5.0",
--- a/src/client.js
+++ b/src/client.js
@@ -1,12 +1,10 @@
-import 'babel-polyfill';
-import { trigger } from 'redial';
-
 import React from 'react';
 import ReactDOM from 'react-dom';
 import Router from 'react-router/lib/Router';
 import match from 'react-router/lib/match';
 import browserHistory from 'react-router/lib/browserHistory';
 import { Provider } from 'react-redux';
+import { trigger } from 'redial';
 import { StyleSheet } from 'aphrodite';
 
 import { configureStore } from './store';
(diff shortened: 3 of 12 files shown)


.babelrc
```diff
@@ -1,4 +1,3 @@
 {
-  "presets": ["es2015", "react", "stage-0"],
-  "plugins": ["transform-react-constant-elements", "transform-react-inline-elements"]
```
justingreenberg: these plugins should only be used in production, moved into prod webpack babel query

justingreenberg:
i added a few more webpack-related enhancements/optimizations:

- update to more performant devtool, since `source-map` generates 1:1 line-column mappings it is much more expensive to generate... in development we can use cheap-module-eval-source-map which only  the maps line numbers from original source
- named async bundle outputs by providing third argument at require.ensure split points
- use babel loader's `cacheDirectory` option to cache transformations to the filesystem, provides ~2x perf improvement
- use transform-runtime with babel-loader to disable per-file runtime injection and remove need to include polyfill in main bundle

next steps would be to configure DLL for vendor deps

jaredpalmer:
Want to work through adding this 0.2

jaredpalmer:
going to evaluate this later today and perhaps carry it into a new PR

justingreenberg:
sounds good, i should be around let me know if you have any questions!

jaredpalmer:
Would things be better if we moved to webpack 2? or is it still not yet ready for primetime?

justingreenberg:
i've been using webpack v2 for about a year—it's definitely stable.. this solution will work fine as implemented since it's instrumenting node's internal module system (it will not work with system loader)...  

some references: 
https://gist.github.com/sokra/27b24881210b56bbaff7 - overview
https://github.com/mxstbr/react-boilerplate/tree/master/internals/webpack - implementation
https://github.com/webpack/webpack/tree/master/examples/harmony-unused - tree shaking 

async reducers/sagas (ssr would require addl hooks on store)
https://github.com/mxstbr/react-boilerplate/commit/c128ee1df6b1e061f4d7da9c28c888f959cd583d

the real benefits of webpack v2 imo are system loader, tree shaking and configuration (base config can return a function that accepts env specific configs)... other than that, everything is pretty much the same

justingreenberg:
i'm not sure why this says ok to merge, looks like it would need to be rebased, so i'm gonna close it... if you want to support just images, here is the gist for for resolving webpack asset names: 

https://gist.github.com/justingreenberg/f1735abffbae3778014f8d14b599a2ef#file-require-patch-js

jaredpalmer:
@justingreenberg thanks man.

PR #63, opened 2016-06-07
use axios instead of custom http client

jaredpalmer:
This PR replaces the custom http client (based on superagent) with the axios and improves existing `callAPIMiddleware` to more closely mirror the middleware found in the [redux real world example](https://github.com/reactjs/redux/blob/ad33fa7314e5db852a306d9475be5cfe22bde180/examples/real-world/middleware/api.js). 

In some respects, this new middleware has more utility than the one in the [redux real world example](https://github.com/reactjs/redux/blob/ad33fa7314e5db852a306d9475be5cfe22bde180/examples/real-world/middleware/api.js) because it 
  1. supports all axios options and http verbs, not just `GET` requests. 
  - can run universally (in node or browser)
  - generates FSA-compliant actions

TODOS:
- [x] replace http client
- [x] use flux standard actions
- [x] update tests
- [x] update docs with examples of symbols


jaredpalmer:
@rowellx68 @justingreenberg @b2whats Thoughts on new middleware?

rowellx68:
Looking good!

Thanks for the docs as well. I think any new major changes should include some for of documentation?

PR #64, opened 2016-06-07
Standard.js

jaredpalmer:
implements standardjs code style, also has a nice helper webpack loader that warns you when you’ve made an error.


jaredpalmer:
fixes #52 as well

PR #65, opened 2016-06-07
use invariant to assert state shape in api middleware

justingreenberg:
saw you merged this already but was just gonna suggest using invariant to assert state shape


jaredpalmer:
LGTM

PR #70, opened 2016-06-08
[WIP] Experimental Workers with RabbitMQ

jaredpalmer:


jaredpalmer:
In order to run end-to-end tests in travis (like server_spec.js), we would have to dockerize everything. Or, we could try to run it all on one instance by modifying the workers. This is what happens with [heroku-node-articles](https://github.com/heroku-examples/node-articles-nlp/blob/master/lib/server.js#L28), the production-level example node app provided by heroku

PR #77, opened 2016-06-10
Fix HMR memory leak in client

justingreenberg:
- moves location variable into render function scope as name conflicts with param return by history.listen
- prevents binding duplicate event listeners to history by unsubscribing before hot reload 

this could also be fixed by adding a dispose handler to module.hot and passing reference to subscription, but this solution seems easier to grok without deep understanding of webpack api


jaredpalmer:
LGTM

Issue #76, opened 2016-06-10
Normalize state


jaredpalmer:
“In a Redux app, state keys like currentPost are usually an anti-pattern.” - Dan Abramov

http://stackoverflow.com/questions/37738268/how-to-avoid-race-conditions-when-fetching-data-with-redux/37738420#37738420

Should implement [normalizr](https://github.com/paularmstrong/normalizr) + selectors

PR #79, opened 2016-06-10
[WIP] Normalizr

jaredpalmer:
An attempt to implement normalizr. `PostList` and `Post` now share schema and a reducer, which is a weird result of this PR.

Still need to think this all the way through, but want to start a discussion.

Attempts to solve #76


wedneyyuri:
@jaredpalmer What do you think about a simple search page in your state? It could make the things more real.

jaredpalmer:
@wedneyyuri trying to find a balance between example, boilerplate, and real-world app. I want things to keep things simple (and easy to delete), yet non-trivial. Ultimately, the point of this project is to show react SSR best-practices.

With that in mind, from a state management perspective, search results and the current `PostList` are identical, but a search input + results list would not leverage SSR and the power of the current config. Furthermore, you would still implement `normalizr` to get rid of the race condition created by storing a `currentPost` or `currentSearchResult` property in the state tree.

jaredpalmer:
TODO:

- [ ] Write tests
- [ ] Decide on folder structure 
- [ ] Add docs to explain why `Post` and `PostList` share a reducer

jaredpalmer:
Honestly, this PR changed my perspective on things.
 \*\*hits joint\*\*

State is just a Database
Selectors are reusable queries

PR #84, opened 2016-06-15
Enable .babelrc plugins only in production

wedneyyuri:
"transform-react-constant-elements" and "transform-react-inline-elements" must not run in development environment

> This transform should be enabled only in production (e.g., just before minifying your code) because although it improves runtime performance, it makes warning messages more cryptic.
https://babeljs.io/docs/plugins/transform-react-inline-elements/
https://babeljs.io/docs/plugins/transform-react-constant-elements/

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,4 +1,8 @@
 {
   "presets": ["es2015", "react", "stage-0"],
-  "plugins": ["transform-react-constant-elements", "transform-react-inline-elements"]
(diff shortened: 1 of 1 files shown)


jaredpalmer:
Good looks

wedneyyuri:
@jaredpalmer Do you have plan about the merge?

Currently, the `propTypes` property is not working.

PR #87, opened 2016-06-18
Provide query to server-side fetch hook

wafflepie:
If you'd want to access query e.g. [here](https://github.com/jaredpalmer/react-production-starter/blob/master/common/routes/PostList/containers/PostList.js#L11), it is undefined. Therefore if you enter `http://localhost:5000/posts?a=1` directly (not by routing to it from a different page), `renderProps.location` has no `query` property. Client-side works fine.

### DIFF
--- a/server/server.js
+++ b/server/server.js
@@ -67,7 +67,7 @@ server.get('*', (req, res) => {
     }
   })
   const routes = createRoutes(store)
-  const history = createMemoryHistory(req.path)
+  const history = createMemoryHistory(req.originalUrl)
   const { dispatch } = store
 
   match({ routes, history}, (err, redirectLocation, renderProps) => {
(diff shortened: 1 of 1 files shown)


jaredpalmer:
cc @markdalgleish

Interesting. Current implementation reflects the example from the [redial](https://github.com/markdalgleish/redial#example-client-usage) docs.

justingreenberg:
i can confirm this behavior, good catch @wafflepie 

however i think a better fix is to instantiate `history` with the full uri path (including queries):

```js
// server.js - line 70
const history = createMemoryHistory(req.originalUrl)
```

jaredpalmer:
@justingreenberg what are pros / cons?

justingreenberg:
this PR just uses express internals to parse the query—which will work—but since we have to build a `history` instance anyway (enhanced with `useQueries` by default) why not use it to parse the query too and maintain functional parity with client

wafflepie:
I agree. I did not realise that `req.path` doesn't include the query. Although the resulting objects should be the same both ways, this is most likely the better way to solve the problem.

PR #134, opened 2016-08-29
Static assets caching using service worker and offline support v1

harrybabu:
1) Cache the webpack static assets using wervice worker for perf
2) Support for offline loading

Using this [Offline pugin]

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,7 +71,8 @@
     "chai-immutable": "^1.5.4",
     "chalk": "^1.1.3",
     "mocha": "^3.0.2",
-    "nyc": "^8.1.0"
+    "nyc": "^8.1.0",
+    "offline-plugin": "^3.4.2"
   },
   "standard": {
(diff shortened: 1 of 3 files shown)


jaredpalmer:
Wow. Super cool. Want to play around and explore a bit!

package.json
```diff
@@ -71,7 +71,8 @@
     "chai-immutable": "^1.5.4",
     "chalk": "^1.1.3",
     "mocha": "^3.0.2",
-    "nyc": "^8.1.0"
+    "nyc": "^8.1.0",
+    "offline-plugin": "^3.4.2"
```
jaredpalmer: Would you be able to move `offline-plugin` from `devDependcies` -> `dependencies` in package.json. Otherwise Heroku won't build properly.

harrybabu: Sure :-)

harrybabu: Done. plz check

PR #139, opened 2016-09-12
babel run time added, to remove duplicate babel helper functions

kishorevarma:
According to
https://github.com/thejameskyle/babel-react-optimize#transform-react-inline-elements
https://babeljs.io/docs/plugins/transform-runtime/


jaredpalmer:
This is having issues with deployment. 😿  . Going to temporarily close and reopen so I can create a review app.

PR #212, opened 2017-03-11
The Next Episode

jaredpalmer:
Pretty much a full rewrite...but soo many goodies to share with everyone!

- [x] **Universal Hotness** (hot reload both server and client during development)
- [x] React Router 4, with @palmerhq's HOC
- [x] Jest
- [x] Fiber
- [x] import() support
- [x] React Helmet
- [x] React SVG Loader
- [ ] Glamor
- [ ] Rapscallion?
- [ ] Express Security (helmet, hpp, csp)
- [ ] babel-plugin-tranform-define for universal config
- [x] File loader, url loader
- [ ] Production builds
- [ ] Docker
- [ ] Throng
- [ ] Docs
- [ ] Document route registration regardless of ssr
- [ ] Recursively resolve promises on nested routes without context shenanigans **Bonus** 
--------
- [ ] frameworky stuff...either merge with backpack or become backpack's first plugin.

PR #213, opened 2017-03-13
Refactored the way redial is triggered client side.

lewis-elliott:
This stops onEnter being called twice when a route is visited.


jaredpalmer:
All efforts are going to #212 at the moment. I'm dropping redial (and probs redux).

PR #255, opened 2017-04-28
Razzle

jaredpalmer:


packages/razzle/scripts/build.js
```diff
@@ -0,0 +1,132 @@
...
+  // Merge with the public folder
+  copyPublicFolder();
+});
+
+function build(previousFileSizes) {
+  // Check if razzle.config.js exists
+  let razzle = {};
+  try {
+    razzle = require(paths.appRazzleConfig);
+  } catch (e) {}
```
ivoreis, line 43: Is there any reason you are swallowing the errors here? Sometime the razzle.config.js exists and isn't being picked and would be nice to log an info message and the stack.

jaredpalmer: `razzle.config.js` is optional

PR #266, opened 2017-05-13
Add Material UI Styled Example

kireerik:
Add [Razzle Material UI Styled Example](https://github.com/kireerik/razzle-material-ui-styled-example) as a submodule to the examples.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "examples/material-ui-styled"]
+	path = examples/material-ui-styled
+	url = https://github.com/kireerik/razzle-material-ui-styled-example


jaredpalmer:
Thanks for this. 

A couple of things:
1. The submodule approach is not going to work with the new `create-razzle-app` #267 that's coming because of how the install from example flag (`--example`) will work.
2. I think this should be split this into two distinct examples: Material UI (`with-material-ui`) and Styled Components (`with-styled-components`).

kireerik:
You are welcome!

1. I see, I can add it directly as well.
2. I see. Basically I have created this example project for my self. So from now on I can start to work on a new project of mine. Which will be a more current version of [Search is Back!](https://searchisback.com/) with more possibilities by the way.
What do you think about adding it as a combination? Like `with-material-ui-with-styled-components` or like `with/material-ui-and-styled-components`.
However in this case the full name would be something like this: `with-material-ui-and-styled-components-and-compression-and-html-minifier`
Note: I am leaving out Express because it is part of the basic example too and it is just called `basic` not `with-express` for example.
In this case we could might shorten the `-and-` parts in the folder name like `with-material-ui--styled-components--compression--html-minifier` or `with-material-ui.styled-components.compression.html-minifier`.
I can split my project, but in this case should I split compression and html-minifier as well or they can be included because they related to Express so it is might considered a basic thing. What do you think?

jaredpalmer:
The styled-components example should only include the code required to use styled component isomorphically. It should not include any other deps. Same goes for MUI.

kireerik:
I see. I will may add these 2 separately then sometime.

PR #269, opened 2017-05-19
Environment-specific and local .env files

jaredpalmer:
Mirror the new multi dotenv setup from CRA...

https://github.com/facebookincubator/create-react-app/pull/1344


jaredpalmer:
At this point it's making more and more sense to add #270 before 0.6.0

PR #282, opened 2017-05-25
Add instructions for yarn create with example

ericclemmons:
It may be worth opening an issue with `yarn` to see if they want `yarn create --whatever --flags--here` to be forwarded to the sub-command, since the ideal is:

```shell
yarn create razzle-app my-app --example with-preact
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,14 @@ by adding  `--example <example-name>` to your command.**
 create-razzle-app --example with-preact my-app
 ```
 
+or with `yarn create`
+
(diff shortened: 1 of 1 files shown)


jaredpalmer:
Good catch!

jaredpalmer:
@ericclemmons hmmm. I think yarn has a few CLI config options (e.g. `yarn init --yes` automatically answers yes to all the questions). That being said, I agree 100% that maybe `yarn create` be treated differently

PR #304, opened 2017-06-22
[Docs] site

jaredpalmer:


jaredpalmer:
Working one formik docs with gatsby, will end up copying that setup to razzle when I’m done

PR #311, opened 2017-07-17
Use npx instead of global install in README

jariz:
`npx` has been introduced in npm 5.2, and I'm guessing it's npm's answer to `yarn create`. ([read the full intro here](https://medium.com/@maybekatz/introducing-npx-an-npm-package-runner-55f7d4bd282b))
It's perfect for tools like `create-razzle-app` which you only run once in a while and would absolutely prefer to be as up to date as possible.
It is fairly new so it might not exist on people's machines in a lot of cases, but then again, so is `yarn create` and it appears to be the suggested way right now to run create-razzle-app with yarn.
But I'm open to adding a warning perhaps telling people to update their npm to 5.2>

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ yarn create razzle-app my-app
 by adding  `--example <example-name>` to your command.**
 
 ```bash
-create-razzle-app --example with-preact my-app
+npx create-razzle-app --example with-preact my-app
 ```
 
 or with `yarn create`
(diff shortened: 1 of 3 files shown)


jaredpalmer:
Good idea. Can you also change the README's of the inner packages too so that the docs match on npmjs.com?

https://github.com/jaredpalmer/razzle/tree/master/packages/create-razzle-app#create-razzle-app
https://github.com/jaredpalmer/razzle/tree/master/packages/razzle

jariz:
Done 👍

PR #312, opened 2017-07-18
Add webpack proxy, react-dev-tools hot client

### LINKED ISSUE #284: Add error overlay like create-react-app's
Create-react-app uses a custom webpack dev server client and its own custom error overlay. We could use switch to [react-dev-utils/webpackHotDevClient.js](https://github.com/facebookincubator/create-react-app/blob/master/packages/react-dev-utils/webpackHotDevClient.js), except that it uses [`window.location.port`](https://github.com/facebookincubator/create-react-app/blob/master/packages/react-dev-utils/webpackHotDevClient.js#L165).  From my initial explorations, this was the dealbreaker. 

That being said, it would be really nice to get an error overlay like CRA's for Razzle

jariz:
😍

screen shot 2017-07-18 at 09 40 38

This fixes #284.
But, this is mostly a POC: it sacrifices a lot, sadly.

Notes:
- I was unable to get this working with `webpack-hot-middleware` & `webpack-dev-middleware`. It appears `react-dev-utils/webpackHotDevClient` uses a couple of hardcoded features only available in `webpack-dev-server` (which is kind of hacky!)  
Sadly, this means we still need to run webpack on a seperate port; my goal was to completely eliminate the need of a separate webpack server, but this makes this pretty much impossible.
- **This breaks backwards compatibility**; it assumes all razzle apps are using the proxy; otherwise the client will be unable to connect.
- This does require all razzle based apps to have `razzle-dev-utils` as a dependency. We could make the webpackProxy part of razzle itself, but it really feels a lot more like a dev util to me.
- It doesn't appear to catch run time errors like create-react-app.  
Haven't been able to figure out why yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ yarn create razzle-app my-app
 by adding  `--example <example-name>` to your command.**
 
 ```bash
-create-razzle-app --example with-preact my-app
+npx create-razzle-app --example with-preact my-app
 ```
 
 or with `yarn create`
--- a/README.md
+++ b/README.md
@@ -49,9 +49,7 @@ Razzle comes with the "battery-pack included" and is part of a complete JavaScri
 ## Quick Start
 
 ```bash
-$ npm i -g create-razzle-app
-
-create-razzle-app my-app
+$ npx create-razzle-app my-app
 cd my-app
 npm start
 ```
@@ -73,21 +71,12 @@ When you’re ready to deploy to production, create a minified bundle with `npm 
 
 ## Getting Started
 
-### Installation
-If you have the latest version of Yarn, you can skip this. Otherwise:
-
-Install Razzle globally:
-
-```
-npm i -g create-razzle-app
-```
-
 ### Creating an app
 
 To create an app, run:
 
 ```bash
-create-razzle-app my-app
+npx create-razzle-app my-app
 ```
 
 or with `yarn create` (new!):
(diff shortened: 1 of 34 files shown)


jariz:
There's also the possibility to builld our own error overlay, but that would bring the extra burden of maintaining that one as well, while other really great solutions like `react-error-overlay` & `react-dev-utils/webpackHotDevClient` exist.

jaredpalmer:
This is pretty amazing work. So thank you x1000000.

Backwards compat is pretty important. Not because I personally care too much, but because Razzle's competition (CRA, Next, react-server, etc.) are really good about maintaining it. Thus, I would explore writing our writing our own overlay based on some of the work you have done, but as a separate package (perhaps called `razzle-dev-tools`). This would be optional, as Razzle is not a runtime dependency right now. 

FYI, if we could figure out how to make Razzle work as simply node.js middleware (like next), I would totally be open to runtime a dep.

jariz:
Yea, I'd be interested in exploring that too. What's really great about next's error handling is that it also shows server side runtime/compile errors, something we probably can never do if we use CRA's error handling, would be really cool if we could add that to razzle as well.

Regarding middleware: I basically had the entire client side compiling working as a middleware, shouldn't be too hard to add server compiling to that as well.
Let's make a separate issue for that though.

jariz:
Going to close this in favor of a from scratch based approach.

Issue #297, opened 2017-06-15
Overriding loader styles in webpack


helguita:
Hello,

I'm trying to override webpack configuration so I can add some options on css-loader and postcss-loader. However I run into an issue that says: 

```
./src/components/mycomponent/MyComponent.css
Module build failed: Syntax Error

(5:1) Unknown word

  3 | // load the styles
  4 | var content = require("!!../../../node_modules/css-loader/index.js?{\"importLoaders\":1,\"modules\":true,\"localIdentName\":\"[name]__[local]___[hash:base64:5]\",\"camelCase\":true}!../../../node_modules/postcss-loader/lib/index.js!./MyComponent.css");

```
after reading some issues over css-loader. It looks like the issue is that I'm overriding the styles that you already defined on razzle conf (see https://stackoverflow.com/questions/37031123/why-am-i-not-being-able-to-compile-sass-with-webpack) and that doesn't go well. I guess that's why Vue created vue-loader.

Given my scenario, should it be best just to fork your project and start from there?

jaredpalmer:
You don't need to fork razzle, but instead totally replace the css loader or add a new one. Here's an example `razzle.config.js` of a project that adds SCSS support.

https://gist.github.com/jaredpalmer/0a91a7bd354b875b913c74f4b16125f7

d4rky-pl:
@jaredpalmer I just spent an hour unsuccessfully trying to add Sass to my example Razzle project until I found this issue and your gist. Might be a good idea to add it to README to save others some time :)

jaredpalmer:
@d4rky-pl Sorry about that. Perhaps an example `with-scss` would be even better?

d4rky-pl:
@jaredpalmer oh yeah, it'd be perfect :) I was actually planning to release this config (probably tomorrow) with some tweaks and configurability as npm package so it can be used in a simliar fashion to `my-razzle-modifictions` from README :)

imjared:
Hey @jaredpalmer, I'm having a similar issue where I'd like to overwrite the default CSS loader. After adding my own loader in `razzle.config.js`, I just end up with dupe CSS loading behavior. A `console.log` of my config.module.rules looks like this (shortened for relevant parts):

```javascript
module.exports = {
  modify: (config, {target, dev}, webpack) => {
    const appConfig = Object.assign({}, config);
   
    appConfig.module.rules.push({
    	test: /\.css$/,
[...]

PR #314, opened 2017-07-24
Add example using Sass

eliot-akira:
This is an example of configuring `razzle.config.js` to support Sass, based on a gist mentioned in #297.


vxna:
Is there any additional configuration required to get it to work with .sass files?

Adding this:

     test: /\.(sass|scss)$/,

Doesn't help and produce an error on compile: 

    ERROR  Failed to compile SERVER with 1 errors
    ./src/styles/App.sass
    Module build failed: Missed semicolon (2:11)

      1 | body
    > 2 |   margin: 0
        |           ^
      3 |   padding: 0
      4 |   font-family: sans-serif
      5 |

vxna:
Sorted, https://github.com/jaredpalmer/razzle/issues/315

sebmor:
If you do 
```
// Comment here
@import 'scsspartial';
```

The build fails with `Unknown word`

It fails on the server

paramaggarwal:
@eliot-akira try using `sass-loader` instead of the `fast-sass-loader` used here. (the original gist in #297 also uses `sass-loader`.)

examples/with-scss/razzle.config.js
```diff
@@ -0,0 +1,73 @@
...
+        fallback: 'style-loader',
+        use: [
+          {
+            loader: 'css-loader',
+            options: {
+              importLoaders: 1,
+            },
+          },
+          postCssLoader,
+          'fast-sass-loader',
```
paramaggarwal: please use `sass-loader` for max compatibility

jaredpalmer: Use `sass-loader` por favor.

eliot-akira: Got it. :)

mhuggins:
~~I've tried using this on my new razzle app, but I'm getting an error:~~

>     Module parse failed: /Users/mhuggins/Development/neo/explorer/src/common/styles/app.scss Unexpected token (1:5)
>     You may need an appropriate loader to handle this file type.
>     | body {
>     |   background-color: blue;
>     | }

My `app.scss` file simply contains this:

```scss
body {
  background-color: blue;
}

.app-component {
  background-color: red;
}
```

~~It seems like my razzle.config.js file isn't even being picked up for some reason.~~

**EDIT**: Turns out I accidentally had the razzle.config.js file in my src path. 😞

examples/with-scss/razzle.config.js
```diff
@@ -0,0 +1,73 @@
+// Based on: https://gist.github.com/jaredpalmer/0a91a7bd354b875b913c74f4b16125f7
+
+const path = require('path')
```
mhuggins: This isn't being used.

paramaggarwal:
This is needed soon - even the `create-react-app` team is realising that default SCSS support is needed by popular demand.

(ref: https://twitter.com/dan_abramov/status/917147209170341889)

examples/with-scss/razzle.config.js
```diff
@@ -0,0 +1,73 @@
...
+      // On the server, we simply use css-loader to deal with scss imports
+      isServer ? 'css-loader' :
+      // For development, include source map
+      dev
+      ? [
+        'style-loader',
+        {
+          loader: 'css-loader',
+          options: {
+            modules: false,
```
paramaggarwal: I just checked and `modules` is `false` by default. So no need. I understand that the original gist from @jaredpalmer had this - but actually not needed.

examples/with-scss/razzle.config.js
```diff
@@ -0,0 +1,73 @@
...
+        'sass-loader',
+      ]
+      // For production, extract CSS
+      : ExtractTextPlugin.extract({
+        fallback: 'style-loader',
+        use: [
+          {
+            loader: 'css-loader',
+            options: {
+              importLoaders: 1,
```
paramaggarwal, line 60: This means that when we `@import` inside an SCSS file, only `post-css-loader` will run on it and not `sass-loader` - is this the correct expectation? I have been trying to understand the implications. I understand that the original gist by @jaredpalmer also has this - but we could add a comment here about what does this actually do.

jaredpalmer: I copied this sass setup from nytimes/kyt should look at their latest release and copy.

mhuggins:
@paramaggarwal Here's an example of a setup that is currently working for me in a production app: https://github.com/neoverse/explorer/blob/master/config/razzle-sass.js

You can see the styles being applied [here](http://explorer.neoverse.io).

jaredpalmer:
@paramaggarwal 

@mhuggins solution is what i've been using in production for 7months, works fine. Confused about what's the issue is?

mhuggins:
Not sure what the issue is, I was just sharing a slightly modified version of this that is working for me.  My solution is derived from this PR with just a little cleanup.

jaredpalmer:
btw you don't even need sass loader on the server. just change css loader's test to include `/.scss$/` and it should be faster.

examples/with-scss/razzle.config.js
paramaggarwal: I read up more on this - what we have here is correct. It basically tells `css-loader` to use `postcss-loader` on all `@import` lines that it comes across in CSS.

mhuggins:
Hmm, I changed it in [this PR](https://github.com/neoverse/explorer/commit/4aac1fe9eebb539d067cbf490ee7ee8a41cb6fd0#diff-080668f2db4ff077b131fc66d6499954), thought I can't remember why.  I do recall I had some kind of issue before adding that.  I'll play around when I have time and add a comment here for why I did so.

paramaggarwal:
All, I think this PR is ready to go and we should merge it in. My comments are just out of curiosity and not as review comments.

@jaredpalmer It gave an error if I use only `css-loader` on the server, so I initially went with using `['css-loader', 'sass-loader']` - but just today, I have changed it to simply use `ignore-loader`.

If you would like to make source maps work for the SASS files, you need to do what I do in the config that I am using: https://gist.github.com/paramaggarwal/e459f177eba98b67c2fa7f26a202d146 which is to give `sourceMap: dev` as an option to all three loaders.

d4rky-pl:
FYI: ignoring CSS/SASS loaders on server side when using CSS modules (manually with `:local`, not on every file with `modules: true`) will result in FOUC (Flash Of Unstyled Content) in React 15 and may result in broken render in React 16 due to difference in generated class names. 

It bit us already in a project once.

jaredpalmer:
Only during dev

Grsmto:
Having the same issue as @sebmor when using @import . Anyone figured out where it comes from?
Seems to kinda work when I delete the declaration and HMR kicks in...

examples/with-scss/razzle.config.js
```diff
@@ -0,0 +1,73 @@
...
+          })
+        ]
+      }
+    }
+
+    appConfig.module.rules.push({
+      test: /.scss$/,
+      use:
+      // On the server, we simply use css-loader to deal with scss imports
+      isServer ? 'css-loader' :
```
Grsmto: Replacing this line by `isServer ? ['css-loader', 'sass-loader'] :` seems to fix the `@import` not working in dev as commented there https://github.com/jaredpalmer/razzle/pull/314#issuecomment-323797303

mhuggins: Yep, this is what I'm doing in my app as well.

https://github.com/neoverse/explorer/blob/4aac1fe9eebb539d067cbf490ee7ee8a41cb6fd0/config/razzle-sass.js#L26

examples/with-scss/razzle.config.js
eliot-akira: OK, removed.

examples/with-scss/razzle.config.js
eliot-akira: OK, removed.

examples/with-scss/razzle.config.js
eliot-akira: Sure, I updated this to use both `css-loader` and `sass-loader` on the server.

Grsmto:
@eliot-akira hey, dev works fine now. But `npm run build` is breaking with `You may need an appropriate loader to handle this file type.`.

eliot-akira:
Just added sourcemaps for all loaders, as per [comment](#issuecomment-339698419) by @paramaggarwal.

@Grsmto Hmm, I'm not seeing that on my end - the build completes as usual.

Grsmto:
@eliot-akira my bad. It's working fine now!

dmitru:
Any updates for this one? What is stopping from merging this PR? 🤔

examples/with-scss/razzle.config.js
```diff
@@ -0,0 +1,77 @@
+// Based on: https://gist.github.com/jaredpalmer/0a91a7bd354b875b913c74f4b16125f7
+
+const autoprefixer = require('autoprefixer')
+const ExtractTextPlugin = require('extract-text-webpack-plugin')
```
hrasoa, line 4: Shouldn't we use [mini-css-extract-plugin](https://github.com/webpack-contrib/mini-css-extract-plugin) ?

examples/with-scss/razzle.config.js
```diff
@@ -0,0 +1,77 @@
...
+              importLoaders: 1,
+            },
+          },
+          postCssLoader,
+          'sass-loader',
+        ],
+      })
+    })
+
+    if (!isServer && !dev) {
```
hrasoa, line 69: If you use [mini-css-extract-plugin](https://github.com/webpack-contrib/mini-css-extract-plugin), you won't need this anymore as ```appConfig.plugins``` already contains the plug in set up:
https://github.com/jaredpalmer/razzle/blob/560a9f490bdcac89b29a03b7d9c471ff21ff6619/packages/razzle/config/createConfig.js#L496

Issue #287, opened 2017-06-02
Node.js 8.0.0+ compatibility issue


kireerik:
I have installed Node.js 8.0.0 under Windows 10 and I am getting the following errors when I am using the `razzle start` command:
```shell
\node_modules\razzle\config\create-config.js:60
        process.env.NODE_PATH.split(path.delimiter).filter(Boolean)
                             ^

TypeError: Cannot read property 'split' of undefined
    at module.exports (\node_modules\razzle\config\create-config.js:60:30)
    at Object.<anonymous> (\node_modules\razzle\scripts\start.js:35:20)
    at Module._compile (module.js:569:30)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:503:32)
    at tryModuleLoad (module.js:466:12)
    at Function.Module._load (module.js:458:3)
    at Function.Module.runMain (module.js:605:10)
    at startup (bootstrap_node.js:158:16)
    at bootstrap_node.js:575:3
\node_modules\execa\index.js:277
                throw (result.error || new Error(result.stderr === '' ? result.stdout : result.stderr));
                ^

Error: null
    at Function.module.exports.sync (\node_modules\execa\index.js:277:26)
    at Object.<anonymous> (\node_modules\razzle\bin\razzle.js:12:9)
    at Module._compile (module.js:569:30)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:503:32)
    at tryModuleLoad (module.js:466:12)
    at Function.Module._load (module.js:458:3)
    at Function.Module.runMain (module.js:605:10)
    at startup (bootstrap_node.js:158:16)
    at bootstrap_node.js:575:3
error Command failed with exit code 1.
```
I am getting a similar error for `razzle build` as well.

With Node.js 7.10.0 it worked and it still works if I install it back.

jaredpalmer:
This line comes directly from https://github.com/facebookincubator/create-react-app/blob/master/packages/react-scripts/config/webpack.config.dev.js#L89

Can you try the latest create-react-app to see if you have the same issue?

kireerik:
I see.

Sure, now I have tested it and create-react-app works with both versions of Node.js.

kireerik:
Also I have used create-react-app as a global module, but I always use razzle as a local module.

jaredpalmer:
I don't have a way to test this out unfortunately. I don't own a Windows PC. 😢

kireerik:
[...]

PR #322, opened 2017-08-21
Fix node path usage

stoikerty:
Hey Jared,

On my exploration through razzle I stumbled over [this compatibility issue](https://github.com/jaredpalmer/razzle/issues/287). The culprit is a race-condition which I fixed by replacing the use of the `process.env.NODE_PATH`-global with an exported variable. Not sure why it only happens on windows, must have something to do with how Node handles imports/exports and envs in that OS.

It was already working like this:
```js
// packages/razzle/config/createConfig.js
// ...
const getEnv = require('./env');

const nodePath = process.env.NODE_PATH;

module.exports = (
// ...
      modules: ['node_modules', paths.appNodeModules].concat(
        nodePath.split(path.delimiter).filter(Boolean)
      )
// ...
````

but I thought that having a dangling variable there would be confusing, therefore I opted to reduce the reliance on the relevant global entirely.

Love the concept of razzle btw, great stuff 👍


jaredpalmer:
Yasss! This is dope.

PR #325, opened 2017-08-24
cleanup webpack style loader and plugin configuration

dtothefp:
#### Description
I wanted to cleanup a little bit of the webpack configuration with some patterns I use in some of my own boilerplates. 
- There was duplication of configuration, i.e. the `css-loader` and `post-css` loader for the `WEB` configuration between development and production.
- The `config.plugins` was being redefined using concat / spread, seems easier to use a `.apply` spread, and also apply this to `config.client.entry`.

#### Issues
- for the `IS_NODE` `css-loader` it has an option of `importLoaders: 1`. Pretty sure this isn't necessary because there are no loaders following it but I left to maintain functionality of previous code.

### DIFF
--- a/packages/razzle/config/createConfig.js
+++ b/packages/razzle/config/createConfig.js
@@ -44,6 +44,50 @@ module.exports = (
   const dotenv = getClientEnv(target, { clearConsole, host, port });
 
   const devServerPort = parseInt(dotenv.raw.PORT, 10) + 1;
+
+  const getStyleLoaders = () => {
+    const styleLoader = require.resolve('style-loader');
+    const cssLoader = {
+      loader: require.resolve('css-loader'),
+      options: {
+        importLoaders: 1,
+      },
+    };
+    const styleLoaders = [
+      styleLoader,
+      cssLoader,
+      {
+        loader: require.resolve('postcss-loader'),
+        options: {
+          ident: 'postcss', // https://webpack.js.org/guides/migrating/#complex-options
+          plugins: () => [
+            require('postcss-flexbugs-fixes'),
+            autoprefixer({
+              browsers: [
+                '>1%',
+                'last 4 versions',
+                'Firefox ESR',
+                'not ie < 9', // React doesn't support IE8 anyway
+              ],
+              flexbox: 'no-2009',
+            }),
+          ],
+        },
+      },
+    ];
+
+    return IS_NODE
+      ? // Style-loader does not work in Node.js without some crazy
+        // magic. Luckily we just need css-loader.
+        [cssLoader]
+      : IS_DEV
+        ? styleLoaders
+        : ExtractTextPlugin.extract({
+            fallback: styleLoader,
+            use: styleLoaders,
+          });
+  };
+
   // This is our base webpack config.
   let config = {
     // Set webpack context to the current command's directory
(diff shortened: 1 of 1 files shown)


packages/razzle/config/createConfig.js
```diff
@@ -237,15 +214,16 @@ module.exports = (
...
-        ...config.plugins,
-        // Add hot module replacement
-        new webpack.HotModuleReplacementPlugin(),
-        // Supress errors to console (we use our own logger)
-        new webpack.NoEmitOnErrorsPlugin(),
-        // Automatically start the server when we are done compiling
-        new StartServerPlugin('server.js'),
-      ];
+      config.plugins.push(
+        ...[
```
jariz, line 218: why put it in an array and destructure it again? why not pass it as the contents as arguments straight away?

packages/razzle/config/createConfig.js
```diff
@@ -261,17 +239,20 @@ module.exports = (
...
+      client: [paths.appClientIndexJs],
+    };
+
     if (IS_DEV) {
       // Setup Webpack Dev Server on port 3001 and
       // specify our client entry point /client/index.js
-      config.entry = {
-        client: [
+      config.entry.client.unshift(
+        ...[
```
jariz, line 250: ?!

jaredpalmer:
So if we end up refactoring this mess (and i agree its a mess). I think we should go with 5 files:

- `webpack.client.dev.js`
- `webpack.client.prod.js`
- `webpack.server.dev.js`
- `webpack.server.prod.js`
- `webpackDevServer.config.js`

@jariz these would just be vanilla objects and NOT functions, so we can one-day write an eject function

jariz:
Yeah, I'd rather implement something like that than continue trying to make sense out of `createConfig` like this PR does.

packages/razzle/config/createConfig.js
dtothefp: @jariz I'm not quite sure I understand your question. from what I understand of spread operator this is the same as

```js
Array.prototype.push.apply(config.plugins, [...<a_bunch_of_plugins>])
```
which is more performant than `.concat` and in my opinion more concise than the pattern of mutating by redefining to the same variable that you are concat'ing into...if that makes any sense 

```js
config.plugins = [
  ...config.plugins,
```

packages/razzle/config/createConfig.js
dtothefp: hot entry needs to go in front of the main javascript entry otherwise the magic doesn't work :sake:

dtothefp:
@jariz @jaredpalmer I think you are on the right track with server vs client config, if you are doing isomorphic this is definitely necessary. but webpack config has many aspects, loaders, plugins, and the base config. this will always be a mess until you break them out accordingly. you should maybe consider something like

```
└── webpack-config
    ├── server
    │   └── base.js
    │   └── loaders.js
    │   └── plugins.js
    ├── client
    │   └── base.js
    │   ├── loaders.js
    │   ├── plugins.js
```

`base.js` might be one level up because it has enough in common between server and client. 

if you would like to go for a refactor and scrap this PR I'd be keen to help out. I like your isomorphic approach because you run both webpack builds at the same time but I'm wondering if for large apps it will become ineffecient because you are compiling CSS twice...i.e. if you are using CSS modules you compile it on client and the server so things don't explode. the approach I've used in the past is to generate detailed stats for the client bundle with something like https://www.npmjs.com/package/webpack-isomorphic-tools and then create a mock CSS loader for the server bundle which just reads the CSS module class names from the JSON

PR #339, opened 2017-09-05
Multicompiler & new logging

### LINKED ISSUE #268: Better logging
Really want improve logging. Seems like it is inevitable that we drop `react-dev-utils/formatWebpackMessages` and use a custom formatter. This would allow us to have one call to 

```js
const multiCompilers = compile([serverConfig, clientConfig])
```
 
In multi-compiler mode, we can still pass the respective `clientCompiler` to `webpack-dev-server` and watch the `serverCompiler` in pretty much the same way. However, we would then be able to write truly universal logger (which would look a lot like `formatWebpackMessages`, but handle multiple configs).

jariz:
Fixes #268 

**Summed up**
- Razzle now uses webpack's multicompile feature, giving us the additional free bonus of seeing server compilation errors from our browser automatically too.
- With the power of [`ink`](https://github.com/vadimdemedes/ink) we now display the validity of both targets at all times, which makes for much quicker glancing.
- Virtual console, managed by [`ink-console`](https://github.com/ForbesLindesay/ink-console). We need this because we need to be able to re-render the status indicators.
- Console only gets cleared before compiling, not after finishing compiling (like before, where you'd possible lose valuable information)

**Decisions made**
- **A new webpack plugin was written to output server files to the build folder**
This had to be done because webpack dev server outputs files to it's own virtual filesystem, which makes sense, but this makes it impossible to launch our server, therefore we still needed a way to write files to disk, of which this seemed to be the best choice.
- **`razzle-dev-utils/webpackHotDevClient.js` was slightly modified** to only accept hot updates if the client hash changes (first 20 chars of the total hash)
- **`razzle-dev-utils/hijackConsole.js` was added to the server entry points.**
In order for the virtual console to work, `ink-console` overrides all `console.*` calls with it's own.
However, `StartServerPlugin` uses the `cluster` node module, which runs the compiled server as a separate process, meaning our custom console calls never get called.
This was solved by injecting `hijackConsole` to the entries which communicates console calls back to the master process (aka razzle) through `process.send`.

**Addendum**
Besides server compilation errors, it'd be really cool if we could get runtime server errors to show up in the browser as well.
I think this PR paves the way for that too!

**Screenshots**
Warning state
screen shot 2017-09-05 at 23 17 55

Error state
screen shot 2017-09-05 at 23 18 25

Loading state
screen shot 2017-09-05 at 23 28 09


jaredpalmer:
Epic. Will try this out either tonight or tomorrow

NickCis:
I'm looking forward in seeing this in Razzle, is there something i can do in order to help?

jariz:
Same here! Still waiting for @jaredpalmer to review.  
I'm seeing it has merge conflicts by now as well too, but I'll probably resolve those together with jared's feedback.

paramaggarwal:
Wow, can't wait for this. Now that the two compilers are combined - can one use https://github.com/FormidableLabs/electron-webpack-dashboard as well?

jariz:
@paramaggarwal Yea, I think so. You're more than welcome to check out my branch and test it out.

jaredpalmer:
@jariz Been procrastinating far too long on this. Cloning and reviewing right meow

jaredpalmer:
Wonder if this has to do with my bash prompt but ora doesn't seem to be working exactly as expected. See screenshot below.

screenshot 2017-11-04 14 36 57

jariz:
Hm, weird.
What shell, term app, etc are you using?

jariz:
@jaredpalmer Should I close this?

paramaggarwal:
Hi, could someone please summarise current situation here. Thanks.

PR #367, opened 2017-10-03
Material Ui Example

NickCis:
This PR adds a Material UI example (it uses the _beta_ branch).

It correctly sets up _jss_ for SSR.


jaredpalmer:
LGTM

PR #377, opened 2017-10-19
Fix "Environment Variables" description

nuragic:
Hello!

First of all, nice project! I'm still playing with it, so I have any other relevant feedback... I can only say good job! The only thing I saw that seems to be incorrect is the one I'm proposing to change with this commit... Actually you can read those variables, so in order to avoid any confusion I'd suggest to change that in the docs :nerd_face: .

Thanks!


nuragic:
Smooth! 😄  Thanks @jaredpalmer

PR #381, opened 2017-10-22
Fix creating multiple server chunks

glennreyes:
When adding Code Splitting with dynamic imports, multiple bundles will be created for the client, but unfortunately for the server too:
image

This PR [disables code splitting](https://medium.com/@glennreyes/how-to-disable-code-splitting-in-webpack-1c0b1754a3c5) by limiting the server chunks to one single file in the node instance:
image

See sample [here](https://github.com/glennreyes/razzle-sample)

### DIFF
--- a/packages/razzle/config/createConfig.js
+++ b/packages/razzle/config/createConfig.js
@@ -259,6 +259,10 @@ module.exports = (
       new webpack.NamedModulesPlugin(),
       // We define environment variables that can be accessed globally in our
       new webpack.DefinePlugin(dotenv.stringified),
+      // Prevent creating multiple chunks for the server
+      new webpack.optimize.LimitChunkCountPlugin({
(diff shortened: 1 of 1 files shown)


glennreyes:
@jaredpalmer Thanks for merging (and of course for this project)! 💪 

Any ETA on releasing this?

robertvansteen:
@glennreyes I tried your sample but it doesn't create any bundles for the client so all the chunks return a 404. This PR won't change that right?

jaredpalmer:
i can cut a new release. 

@glennreyes would you be willing to submit an PR with an example with react-loadable?

glennreyes:
@rovansteen I just rebuilt and looked ok?
image

@jaredpalmer Sure, will do!

jaredpalmer:
$WAG!

PR #410, opened 2017-11-23
Add RAZZLE_ENV_PREFIX

### LINKED ISSUE #394: Allow custom prefix for environment variables
Right now razzle forces you to use `RAZZLE_` prefix, it would be nice to be able to replace this with custom one. [Looking at code, it seems that the change would be rather simple](https://github.com/jaredpalmer/razzle/blob/master/packages/razzle/config/env.js#L54), let me know if you're OK with this and I'll send a PR :)

d4rky-pl:
Fixes #394


panayi:
Any update on this?

jaredpalmer:
Is this still needed? Razzle now supports runtime variables in env

krazyjakee:
This is no longer needed

Issue #397, opened 2017-11-16
Check for port in use when running npm run start to avoid EADDRINUSE


jolyonruss:
If I have another application running on port 3000 when I run `npm run start` I get the classic `EADDRINUSE` error.

Steps to reproduce:
 - have another node instance running on port 3000
 - create a fresh razzle project and run `npm run start`

Suggestions, follow the same pattern as create-react-app and check if the port is in use at startup

jaredpalmer:
Yeah this would be a nice feature

jolyonruss:
I assume that could be implemented around line 80 of `razzle/scripts/start.js`?

guyb7:
Just to note that we actually need 2 ports: one for the app and one for the dev-server.
I can work on this. Welcoming any ideas/suggestions 🙂

jolyonruss:
Sounds good @guyb7, that's something I hadn't considered

PR #414, opened 2017-11-29
Check if ports in use in start script (#397)

guyb7:
Using the `choosePort` function in `react-dev-utils` as it's already a dependency.
I wrapped the start script in a `main` function, as `choosePort` returns a promise.

Also, for the dev-server port I introduced a new environment variable `PORT_DEV` to allow providing non sequential ports.


jariz:
lgtm

PR #418, opened 2017-12-04
Fix react loadable example

tkloht:
I noticed some problems with the react-loadable example.

First, when running the dev-server:
react-loadable tracks which chunks were required for the server-render, then we insert script-tags for these chunks. The problem when running on the dev-server is that these have to be loaded from `localhost:3001/static/js/1.chunk.js` (for example), instead of  just loading by relative path from `static/js/1.chunk.js`. With the current configuration the dev-server just returns index.html, instead of the js chunks, resulting in a parse error of course ("Uncaught SyntaxError: Unexpected token <")
I have simply changed the source of the script-tags for preloading to a absolute path to `http://${HOST}:${PORT}/${chunkname}`. If there is a better way to make this work please let me know :)

Second, for production:
The chunks were not loaded in the correct order.
We need to load the main bundle before the chunks, because the main bundle has all the "webpack-stuff" for loading the chunks. Loading the chunks before the main bundle results in error "Uncaught ReferenceError: webpackJsonp is not defined".
With the current setup the main bundle is loaded with defer, while the chunks are loaded without defer so they are loaded first, leading to the error.

I have removed the defer from the main bundle, this also reflects how the scripts are loaded in the react-loadable documentation.
I think it might also be possible to set the chunks to deferred loading, but:
- I am not sure if the order for loading the scripts is guaranteed when using defer. It seems to work as expected in chrome, i.e. deferred scripts load in the same order they are listed in the html. I have no idea if this is just coincidence  or how other browsers behave though.
- the react-loadable docs recommend to defer the hydration until all chunks are preloaded, see [here](https://github.com/thejameskyle/react-loadable/blob/f1bb24750ecc47864df8355f2ae69b42ea42035a/README.md#waiting-to-render-on-the-client-until-all-the-bundles-are-loaded) (also added in this PR). To my knowledge it is not possible to detect that all the chunks are loaded when using deferred loading, because you would use an inline script and that cannot run deferred. It would be possible to wait for all deferred scripts, but then you would also have to wait for all other scripts that you may be using.

I hope this makes sense, it is meant more as a bug report and call-for-help because i'm not entirely sure if there are better ways to fix these issues.

### DIFF
--- a/examples/with-react-loadable/src/client.js
+++ b/examples/with-react-loadable/src/client.js
@@ -4,14 +4,16 @@ import { hydrate } from 'react-dom';
 import Loadable from 'react-loadable';
 import BrowserRouter from 'react-router-dom/BrowserRouter';
 
-Loadable.preloadReady().then(() => {
-  hydrate(
-    <BrowserRouter>
-      <App />
-    </BrowserRouter>,
-    document.getElementById('root')
-  );
-});
+window.main = () => {
+  Loadable.preloadReady().then(() => {
+    hydrate(
+      <BrowserRouter>
+        <App />
+      </BrowserRouter>,
+      document.getElementById('root')
+    );
+  });
+}
 
 if (module.hot) {
   module.hot.accept();
--- a/examples/with-react-loadable/src/server.js
+++ b/examples/with-react-loadable/src/server.js
@@ -45,9 +45,10 @@ server
   <body>
     <div id="root">${markup}</div>
     ${process.env.NODE_ENV === 'production'
-      ? `<script src="${assets.client.js}" defer></script>`
-      : `<script src="${assets.client.js}" defer crossorigin></script>`}
+      ? `<script src="${assets.client.js}"></script>`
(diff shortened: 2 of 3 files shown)


jaredpalmer:
cc @thejameskyle 

How does react-loadable work or not work with `<script defer>`?

jamiebuilds:
All the scripts need to be loaded before you call `window.main()`, it can work with `defer` if you somehow wait for them all to load.

jamiebuilds:
Although I also would expect it to work as long as the main bundle is loaded ahead of time.

tkloht:
do you mean just setting the main bundle and all chunks to defer? I tried that and it seems to work, although i don't quite understand why - wouldn't the main bundle execute before all the chunks are loaded? Or does the Loadable.preloadReady() take care of that?

tkloht:
* forgot to say, I also removed the window.main() and just call Loadable.preloadReady directly

jamiebuilds:
I don't think that Webpack does anything to guarantee the order of operations when loading bundles through script tags.

Issue #409, opened 2017-11-23
README mismatch


d4rky-pl:
There seems to be a growing mismatch between README from razzle repository and README in `packages/create-razzle-app/templates/default/README.md`.

I propose to either sync them automatically or rewrite `create-razzle-app` README into something shorter that says few words about what RAZZLE is (or gives some pointers on what should be in application's README) and then links to existing documentation on GitHub.

jariz:
How about a soft link? Or does that not work well together with npm?

jaredpalmer:
We should probably write a precommit task that copies the README in a lot of places.

PR #427, opened 2017-12-14
Issue #365 - Added option to override package manager.

### LINKED ISSUE #365: Add option to choose a different package manager
We should give developers the freedom to choose their own package manager, rather than always use yarn if it's available:

https://github.com/jaredpalmer/razzle/blob/master/packages/create-razzle-app/lib/utils/get-install-cmd.js#L14

The only solution right now to this is globally removing yarn - which is not really a solution; I want to be able to keep supporting projects that use yarn, while favoring npm for new projects.
Preferably, we'd add a flag to `create-razzle-app` in which we can give a preference.

RobinTo:
Implemented parameter for create-razzle-app to override package manager, could resolve #365 if useable.

`create-razzle-app my-project -p npm`

Command|Info|Valid values
--- | --- | ---
`-p --packageManager`|Override package manager, default to yarn.|`npm`, `yarn`

Tested this on mac using
* `npx ./razzle/packages/create-razzle-app/bin/create-razzle-app test-razzle-project -p npm`
* `cd test-razzle-project`
* `npm start`

If the option is not supplied it defaults to old behaviour, otherwise checks that it is one of `yarn` or `npm`, and defaults to old behaviour if not. Could remove check for `yarn` or `npm` if you want to enable any value to be used.

> I want to be able to keep supporting projects that use yarn, while favoring npm for new projects.

Due to the sentence above I was unsure wether to set the default to npm, but currently it default to yarn as previously done. Could be changed if desired.

I hope you find this useful, even though the issue was marked as low priority. 

Some yarn lockfiles unrelated to my modifications appear to have many changes. I'm not very familiar with the inner workings of yarn, but I suspect it might be due to newer yarn version when I ran the initial yarn setup?


jariz:
I'm slowly seeing the JS hemisphere moving back to npm, but correct me if I'm wrong.
Eitherway, yarn was the previous default so probably best to keep it that way.

jariz:
@RobinTo Can you update all readme's? There are more than 2.
I know this is kind of annoying but it's something that's being worked on (#409)

RobinTo:
Of course, all the relevant readme files should now be updated.

packages/create-razzle-app/lib/messages.js
```diff
@@ -22,6 +22,10 @@ exports.exampleHelp = function() {
   )}`;
 };
 
+exports.packageManagerHelp = function() {
+  return `Wich package manager (npm or yarn) to utilize for setup.`;
```
jaredpalmer: Which***

jaredpalmer (commented):
Not sure why lockfiles are in this commit. May need to open a new PR

jaredpalmer:
Is there a way to automatically tell based on whether user has called this with `npx` or `yarn create`?

RobinTo:
I fixed the spelling error. Let me know if you want me to open a new pull request where I reset the lockfiles to the versions present on current master.

As far as I could see with some light searching, there is no way to detect if it was done through yarn create or directly, since according to https://yarnpkg.com/lang/en/docs/cli/create/ yarn create is equal to

```
$ yarn global add create-react-app
$ create-react-app my-app
```

where the end result is the same as if you had installed create-react-app globally with any packagemanager and ran the command directly. If you have any suggestions where to look for making it work, please let me know though and I'd be happy to look into it further.

jariz:
I've done a little research by running `yarn create` and sadly [`process.argv`](https://nodejs.org/api/process.html#process_process_argv) returns the same as it would when directly running the script.
On top of that [`process.ppid`](https://nodejs.org/api/process.html#process_process_ppid) is set to undefined, so no way to get information about who started the script either.

So, short answer to @jaredpalmer's question: Nope.

jariz:
Scratch that, ppid only exists in 9.2+ and it does appear to return the yarn pid, but then again, that version has only been out for 2 months.
So, I think this is kind of out of scope for now, unless we're willing to do all sorts of [crazy platform specific hacks](https://stackoverflow.com/questions/5566009/node-js-parent-process-id).

jaredpalmer (commented):
Nice

PR #436, opened 2017-12-21
Improve TS example

jariz:
This PR attempts to bring the TS example as much in line with the default template as possible.

It:
- Removes eslint (it's of no use w/typescript)
- Enables tslint by default because eslint is enabled by default too.
- Adds react router and all the default routes & assets.
- Don't use `import * as Foo` (`allowSyntheticDefaultImports` is enabled so no need for that)
- Fixes incorrect line numbers in razzle.config.js


jaredpalmer:
@jariz Should we make this more real world out of the box? I use cache-loader, thread-loader, and happyPackMode + Fork TS Check plugin. 


```js
const ForkTsCheckerWebpackPlugin = require('fork-ts-checker-webpack-plugin');

...
 // Declare our TypeScript loader configuration
    const tsLoader = {
      include,
      test: /\.tsx?$/,
      use: [
        { loader: 'cache-loader' },
        {
          loader: 'thread-loader',
          options: {
            // there should be 1 cpu for the fork-ts-checker-webpack-plugin
            workers: 2,
          },
        },
        {
          loader: 'ts-loader',
          options: {
            happyPackMode: true, // IMPORTANT! use happyPackMode mode to speed-up compilation and reduce errors reported to webpack
          },
        },
      ],
    };

    // Fully replace babel-loader with ts-loader
    config.module.rules[babelLoader] = tsLoader;

    if (target === 'web') {
      config.plugins.push(
        new ForkTsCheckerWebpackPlugin({
          checkSyntacticErrors: true,
          formatter: 'codeframe',
          tslint: './tslint.json',
          watch: './src',
          workers: 2,
        })
      );
    }
```

jariz:
I'm all for improving performance wherever possible, but we'd probably not want to make the razzle.config.js too advanced, it might become a bit overwhelming otherwise (which would kind of defeat the purpose of razzle).
I've used at-loader before with backpack and it seemed to do multicore compilation out of the box, so that would be my preferred way to go.

PR #457, opened 2018-01-16
Add very basic hooks before and after compile, allow paths to be modified

jaredpalmer:
This is a REALLY quick and dirty bandaid for a larger plugin system I have planned.

## Pros

- Can move After.js to Razzle after this

## Cons

- Not a real plugin system
- Not composable at all

Related https://github.com/jaredpalmer/after.js/pull/42


yunlong-tang:
When this PR can be merged? Now the default paths become a block to me. I need to change to the output path since I already use the 'build' for other things.

jaredpalmer:
Closed in favor of #605

PR #461, opened 2018-01-18
Finish up css-modules integration

b-m-f:
uses locals on server, adds comments and updates readme


b-m-f:
@jaredpalmer could you review this? Without this the css modules wont work properly on first render.

PR #462, opened 2018-01-18
Start migrating to babel 7.x

JacopKane:
Migration of babel 7.x in order to publish changes under a `next` or `beta` tag


examples/basic/package.json
```diff
@@ -14,7 +14,7 @@
     "react-dom": "^16.0.0"
   },
   "devDependencies": {
-    "raf": "^3.3.2",
+    "raf": "^3.4.0",
```
jaredpalmer, line 17: We should remove raf from all example and include our config/polyfill in createJestConfig like CRA does.

packages/babel-preset-razzle/dependencies.js
```diff
@@ -0,0 +1,57 @@
+/**
+ * Copyright (c) 2015-present, Facebook, Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+'use strict';
+
+module.exports = function(api, opts) {
```
jaredpalmer: I think we ought to just ship with babel-preset-react-app going forward. Some of the issues I had early on with razzle appear to be resolved (HMR would break). We should probs just export it for continuity and backwards compat. No need to copy and paste I think.

jaredpalmer:
This looks awesome. I will test this out later tonight.

packages/babel-preset-razzle/dependencies.js
JacopKane: Yes, that would be a much better alternative to also reduce the maintenance

JacopKane:
Cheers, I published some of the modules with `-babel-7` suffix on npm to give them a shot. Will try them further later tonight. Right now I can say jest and build started working couldn't try all the parts yet

packages/razzle/scripts/test.js
```diff
@@ -28,20 +26,26 @@ require('../config/env');
 const jest = require('jest');
 const argv = process.argv.slice(2);
 
-// Watch unless on CI or in coverage mode
-if (!process.env.CI && argv.indexOf('--coverage') < 0) {
+// Watch unless on CI, in coverage mode, or explicitly running all tests
+if (
+  !process.env.CI &&
+  argv.indexOf('--coverage') === -1 &&
```
olehreznichenko, line 32: More readable to use `String.prototype.includes` ?

JacopKane: That code is introduced by this commit actually: https://github.com/facebookincubator/create-react-app/commit/19e0bb1881e24fb1ee3fe421413d4d87e67f68dd

Maybe to avoid ES6 usage not sure but you are right

JacopKane:
Do you think we can implement something better than copy and paste without losing original commits of `create-react-app`? Maybe figure out the things we are using without modification and import them?

jaredpalmer:
Should we wait for CRA 2 or go ahead and push this out in a beta tag?

JacopKane:
I've had some minor issues when I check last time. Sorry I've been quite busy with other stuff. Soon I'll revisit this since I also need to use it.

In the meantime, CRA team also released the changes they made for babel-7 so this means a whole upgrade to build tools and utilities in addition to places that I changed.

PR #471, opened 2018-01-22
Adds Ant Design example (with custom theme).

eduludi:
Example with [Ant Design](https://ant.design) that includes the webpack configuration to override theme Less variables (using [`less-loader`](https://github.com/webpack-contrib/less-loader)).


jonjamz:
The package.json seems to be missing the Extract Text Plugin.

jonjamz:
To get this working with the latest of everything I had to change from `'static/css/[name].[contenthash].css'` to `'static/css/[name].[md5:contenthash:hex:20].css'`. Ideally it would use the `mini-css-extract-plugin` though. Really great work, I spent more than a day trying to make this work before I found your repo.

cbbfcd:
not work~~~~

Issue #512, opened 2018-02-14
Interested in a updated reason example?


ulrikstrid:
So I've tinkered with razzle and reason and even got routing working (client and server) with the new built in ReasonReact router.
Want me to clean it up and make a PR?

jaredpalmer:
Yes please!

ulrikstrid:
Will do in the next couple of days then 👍

ulrikstrid:
This should have been closed as my PR was merged

PR #513, opened 2018-02-14
Update reason example

ulrikstrid:
Following up on #512

### DIFF
--- a/examples/with-reason-react/src/client.js
+++ b/examples/with-reason-react/src/client.js
@@ -3,7 +3,7 @@ import './client.css';
 import React from 'react';
 import { render } from 'react-dom';
 
-const App = require('../lib/js/src/app').comp; // BuckleScript output directory
+import App from '../lib/js/src/App'; // BuckleScript output directory
 
 render(
   <App title="Welcome to Razzle Reason React" />,
--- a/examples/with-reason-react/src/server.js
+++ b/examples/with-reason-react/src/server.js
@@ -2,7 +2,7 @@ import React from 'react';
 import express from 'express';
 import { renderToString } from 'react-dom/server';
 
-const App = require('../lib/js/src/app').comp; // BuckleScript output directory
+import App from '../lib/js/src/App'; // BuckleScript output directory
 
 const assets = require(process.env.RAZZLE_ASSETS_MANIFEST);
 
@@ -13,7 +13,7 @@ server
   .use(express.static(process.env.RAZZLE_PUBLIC_DIR))
   .get('/*', (req, res) => {
     const markup = renderToString(
-      <App title="Welcome to Razzle Reason React" />
+      <App title="Welcome to Razzle Reason React" initialUrl={req.url.substring(1)} />
     );
     res.send(
       `<!doctype html>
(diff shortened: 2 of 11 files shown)


examples/with-reason-react/src/server.js
```diff
@@ -13,7 +13,7 @@ server
   .use(express.static(process.env.RAZZLE_PUBLIC_DIR))
   .get('/*', (req, res) => {
     const markup = renderToString(
-      <App title="Welcome to Razzle Reason React" />
+      <App title="Welcome to Razzle Reason React" initialUrl={req.url.substring(1)} />
```
ulrikstrid: This is kind of a hack to get rid of the initial "/", should I change it to something else?

examples/with-reason-react/src/server.js
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import express from 'express';
 import { renderToString } from 'react-dom/server';
 
-const App = require('../lib/js/src/app').comp; // BuckleScript output directory
+import App from '../lib/js/src/app'; // BuckleScript output directory
```
ulrikstrid: We can change `"in-source"` in  bsconfig.json to `true` and just require it like normal, `import App from './App'`

ulrikstrid:
The Router and Link components are borrowed from https://github.com/thangngoc89/reason-react-router with slight modifications to allow passing in a URL for ssr.

examples/with-reason-react/src/client.js
```diff
@@ -3,7 +3,7 @@ import './client.css';
 import React from 'react';
 import { render } from 'react-dom';
 
-const App = require('../lib/js/src/app').comp; // BuckleScript output directory
+import App from '../lib/js/src/App'; // BuckleScript output directory
```
ulrikstrid, line 6: We can change "in-source" in bsconfig.json to true and just require it like normal, `import App from './App'`

jaredpalmer:
> The Router and Link components are borrowed from https://github.com/thangngoc89/reason-react-router with slight modifications to allow passing in a URL for ssr.

Does the built-in reason router support SSR?

ulrikstrid:
If you take a look at the Router component there is nothing that needs the browser there. The only thing that is needed is for us to pass a initial URL because then we will use that instead of dangerously getting the URL.
So on the server we really don't use the router, we just use the switch statement in App.re.

drejohnson:
I've just noticed that going to root url actually renders "page not found" instead of home component. Going to "/home" or "/counter" rendered correctly on the server.

ulrikstrid:
Noticed that now as well. In my app I added this util:
```ocaml
/* copied from  ReasonReact */
let urlToUrlList = url =>
  switch url {
  | ""
  | "/" => []
  | _ =>
    /* remove the preceeding /, which every pathname seems to have */
    let raw = Js.String.sliceToEnd(~from=1, url);
    /* remove the trailing /, which some pathnames might have. Ugh */
    let raw =
      switch (Js.String.get(raw, Js.String.length(raw) - 1)) {
      | "/" => Js.String.slice(~from=0, ~to_=-1, raw)
      | _ => raw
      };
    raw |> Js.String.split("/") |> Array.to_list;
  };
```
And then parse the URL with that (this also solved that I had to remove the leading "/"). Not sure if that is too much to add for this simple example or if I should include it. Would love for  ReasonReact to expose it instead.

drejohnson:
I would add this to the example for now but it would be preferable if ReasonReact exposed something like this.

ulrikstrid:
@jaredpalmer I think this is fine. Anything else you want me to fix?

PR #530, opened 2018-03-08
Add CLIENT_PUBLIC_PATH env var

gregmartyn:
I'm running this in docker. On my host, I connect to the container using a hostname like myproject.docker. Inside the container, the app has to listen on 0.0.0.0.

Currently, if I set the HOST env var, it tries to listen on a non-existent host:

```
Error: Could not find an open port at myproject.docker.
Network error message: getaddrinfo ENOTFOUND myproject.docker

    at detect.then.err (/var/www/html/razzle/packages/razzle/node_modules/react-dev-utils/WebpackDevServerUtils.js:420:13)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:118:7)
```

By adding this CLIENT_PUBLIC_PATH env var, I can leave HOST at the default, and use the new env var to tell the client where to connect.


gregmartyn:
@jaredpalmer could you reopen this? I'm hoping to get it merged in, and would be happy to make any changes necessary.

gregmartyn:
Reopened as #740

PR #547, opened 2018-03-21
Patching for IE11

marklearst:
Mirroring the patch for react-dev-utils that causes dev utils to break in IE11. strip-ansi package arrow function is causing this issue. See react-dev-utils patch PR that was accepted to fix this issue: https://github.com/facebook/create-react-app/pull/2692


jschertz:
👍

brunorzn:
That's great. Am I wrong or this fix is still not present in the latest razzle 0.8.14 npm package ?

PR #549, opened 2018-03-22
Add an example with Hyperapp

eschaefer:
This adds an example using [Hyperapp](https://github.com/hyperapp/hyperapp), which is heavily inspired by the Elm Architecture.

The example includes a minimal counter (+/-) since the flow of state/actions/view is a core concept.

Let me know if I can make anything clearer or simpler!


examples/with-preact/src/server.js
```diff
@@ -21,12 +21,16 @@ server
...
+        ${
+          assets.client.css
+            ? `<link rel="stylesheet" href="${assets.client.css}">`
+            : ''
+        }
+        ${
+          process.env.NODE_ENV === 'production'
+            ? `<script src="${assets.client.js}" defer></script>`
+            : `<script src="${assets.client.js}" defer crossorigin></script>`
+        }
```
eschaefer, line 33: Didn't mean to change this file, sorry. Just some autoformat from Prettier.

PR #556, opened 2018-03-26
Added the possibility for Build to use NODE_ENV before defaulting

fede:
NODE_ENV was hardcoded to 'production'. This made it impossible to set staging, integration or any other NODE_ENV.

### DIFF
--- a/packages/razzle/scripts/build.js
+++ b/packages/razzle/scripts/build.js
@@ -1,7 +1,7 @@
 #! /usr/bin/env node
 'use strict';
 // Do this as the first thing so that any code reading it knows the right env.
-process.env.NODE_ENV = 'production';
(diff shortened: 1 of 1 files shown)


annez:
I completely agree with this change, builds shouldn't be hardcoded to any sort of environment because people choose to name things differently, and in most environments you will have a staging as well. Unless @jaredpalmer you suggest doing this another way for, say, a URL config map of API endpoints?

We have things like:
```json
{
   "production": "https://foo",
   "staging": "https://bar",
   "default": "localhost"
}
```

These URLs are shared across both client and server, and currently we cannot use the same environment variable to generate the right builds before running this.

jaredpalmer:
Closed in favor of #611

PR #570, opened 2018-04-07
add 404 page and test example to with-afterjs

krazyjakee:
Seemed like an error page and a working test may as well be included by default.
Hope you agree :)


krazyjakee:
I still feel this is relevant

PR #575, opened 2018-04-10
#fix starting server code after compile assets (sync)

ipostol:


chaffeqa:
I can confirm this fixes #578 for me 👍

chaffeqa:
Thanks Jared!  Love your work!

jaredpalmer:
Thanks! I think I may have a slightly faster solution that involves seeding assets.json with a default setup and reading webpack config. However it would still create a race condition. The unfortunate thing about this PR is that it slows down startup time by ~2x.

chaffeqa:
ah yea that makes sense, however I totally agree that x2 perf hit is worth just making it work for now, then optimizing it later.

I'm oping to have the bandwidth to help in this with PR's, but in the meantime really appreciate all you are doing here!

jaredpalmer:
Thanks. Definitely appreciate all contributors!

PR #576, opened 2018-04-10
add emitFile in url-loader options

dunhuang:
There are 2 'static/media' directories after 'npm run build': 
-  build/public/static/media,  created by client side's build
-  build/static/media,  created by server side's build

Only the client side media path is useful.
I added 'emitFile' option to avoid server side media directory generated.


jaredpalmer:
Nice! Was wondering why that was happening.

PR #585, opened 2018-04-19
Add typescript after.js example

worudso:
It's `with-afterjs` + `with-typescript`.
Everything is working ok except `typings/index.d.ts`.
`typings/index.d.ts` seems to be ignored, then `react.svg` and `assets.json` cannot be recognized as typescript module(module declaration for "*.json" should be added). So the logo is not shown now. I think I made a stupid mistake but I can't find it. 
Also the type of `After` component is wrong. It doesn't allow `data` props. I tried to hack it and the conclusion is that it's because of `withRouter` component. I think you should force an explicit type for `After` than leaving it inferred by compiler(or fix `withRouter` type?).


worudso:
I fixed logo bug. Now the type of `After` is the only problem.

worudso:
https://github.com/jaredpalmer/after.js/issues/97 solves the `any` typing problem. I think everything is ok now.

jaredpalmer:
Please remove package-lock.json's in another commit

worudso:
No package-lock.json now.

corydeppen:
Will this be merged soon, or are there things left to address?

smkhalsa:
@jaredpalmer anything else holding this up? I'd also like to see this merged.

PR #589, opened 2018-04-24
Upgrade typescript example

HofmannZ:
With the PR we have access to the new fragment syntax in TypeScript. It also upgrades all TypeScript dependencies to their latest version. 

I've personally been running this updated config for a week now, and the compile time has improved load with new TypeScript version 🚀.


jaredpalmer (approved):
Lgtm

PR #593, opened 2018-04-27
Better defaults for Typescript example

chaffeqa:
After running into multiple bugs having to deal with whether typescript was precompiling the same way that the babel version was (some include: jest testing, repl loading, mjs support) I realized that the big issue is that razzle had figured out solutions for each env based on webpack, so instead of having the added complexity of not knowing if I had proper typescript parody, I figured just remove that layer and use typescript for the one thing I wanted: type checking.

Happy to go into detail on those bugs, but I think the main point is this approach seems a better DX experiance out of the box, while offering more heavy solutions as comments.

### DIFF
--- a/examples/with-typescript/razzle.config.js
+++ b/examples/with-typescript/razzle.config.js
@@ -62,7 +62,7 @@ module.exports = {
     // then do the following:
     //
     // - COMMENT out line 55
-    // - UNCOMMENT line 62
+    // - UNCOMMENT line 67
     //
     // config.module.rules[babelLoader] = tsLoader;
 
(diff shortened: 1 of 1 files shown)


examples/with-typescript/razzle.config.js
```diff
@@ -8,8 +8,8 @@ module.exports = {
       '.ts',
       '.tsx',
     ]);
+ 
 
-    config.devtool = 'cheap-module-source-map';
```
jaredpalmer, line 12: Think this will break HMR if you remove.

chaffeqa: I think thats the default anyways right?

https://github.com/jaredpalmer/razzle/blob/next/packages/razzle/config/createConfig.js#L93

jaredpalmer: Good looks.

PR #601, opened 2018-05-03
Remove raf from examples

lucasterra:
* Removes `raf` from all examples, since it's already included in `razzle` itself.
* Adds a `.gitignore` with `package-lock.json` and `yarn.lock` to the examples folder. This will prevent people from submitting those for new examples.


jaredpalmer:
LGTM

Issue #437, opened 2017-12-21
RFC: Gatsby-like Plugin System


jaredpalmer:
We could start publishing `razzle-plugin-xxxxxx`

```js

module.exports = {
   plugins: [
     'typescript',   // will execute razzle-plugin-typescript in if installed in package.json, will warn otherwise
     'scss',
     'heroku',
     'preact',
     'offline'     
  ],
  modify() {
    // ...same as before but will be called AFTER plugins have run
  }
}

jariz:
Oh I love this. It would also solve the problem I mentioned earlier about not making the TS config too complex.

lucasterra:
@jaredpalmer mind if I implement this? Or you are already working on it?

jaredpalmer:
Go for it!

PR #605, opened 2018-05-05
Razzle plugins + razzle-typescript-plugin

lucasterra:
- [x] gatsby-like plugin system (#437).
- [x] `razzle-plugin-typescript` (tests included)
- [x] Change TypeScript example to use the newly created plugin

---------------------------

It works like this:

Using the plugin with the default options

```js
// razzle.config.js

module.exports = {
  plugins: [
    'typescript'
  ],
};
```

Passing custom options to the plugin:

```js
// razzle.config.js

module.exports = {
  plugins: [
    {
      name: 'typescript',
      options: {
        useBabel: true,
        tsLoader: {
          transpileOnly: true,
          experimentalWatchApi: true,
        },
        forkTsChecker: {
          tsconfig: './tsconfig.json',
          tslint: './tslint.json',
          watch: './src',
          typeCheck: true,
        }
      },
    },
  ],
};
```


lucasterra:
Ideally, a plugin should also be able to change the `createJestConfig(...)`, but I will leave it for another PR.

jaredpalmer:
This looks epic. On my phone right meow, but will dig into the code tomorrow. I have some ideas about the TypeScript plugin (and different default optimizations) which i’ll tack on. I agree about jest config, we def need a modifyJest() and the TypeScript plugin should have ts-jest setup. 


cc @jorilallo

jaredpalmer:
As for the official TS plugin, lets get rid of  tslintloader. We should put tsloader into transpile only and experimentalWatchApi. Then add fork-ts-checker-webpack-plugin

lucasterra:
Removed `tslint-loader`.
Added `experimentalWatchApi`, `transpileOnly` and `fork-ts-checker-webpack-plugin`.

Should we add `thread-loader`/`cache-loader` to `ts-loader` too?

chaffeqa:
I have been brooding on this subject the last week, I'd like to offer this alternative idea:

Babel 7 has added [babel/preset-typescript](https://www.npmjs.com/package/@babel/preset-typescript).  Basically it just strips the typescript while compiling down.

With that in mind, I really like the idea of giving full compile control back to babel, and use typescript for linting / DX.

FYI the issues I've been running into include:
* Trying to run 1-off scripts locally (needs something like: `"repl": "node -r esm --require ts-node/register/transpile-only ./bin/repl"`)
* Trying to run 1-off scripts on server (same file, but don't want to add typescript to `dependencies` and have to bundle it in a deploy 👎 )
* Fix jest to support `after.js` dynamic imports (bombs out due to something with typescript --> babel)
* Issues with `.mjs` (may not be related, but the fact that I'm using x2 compilers just complicates so much as far as possible issues introduced)

I'm hoping to fork razzle at some point to attempt to try the idea (need to fork to update the babel razzle preset) and will get back to you about the experiment.  Just thought I'd add it here FYI.

jaredpalmer:
I actually found perf degradations when using thread and cache loader with the changes. So my gut says to leave them out.

jaredpalmer:
We need a full typescript plugin. We may also do a babel-typescript plugin for Babel 7 TS. As you may or may not know, not all features of TS are supported in Babel TS. So it makes sense to split them I think.

examples/with-typescript/src/App.tsx
```diff
@@ -1,4 +1,4 @@
-import * as React from 'react';
```
jaredpalmer, line 1: Put this back?

examples/with-typescript/src/client.tsx
```diff
@@ -1,10 +1,10 @@
-import * as React from 'react';
-import * as ReactDOM from 'react-dom';
+import React from 'react';
```
jaredpalmer, line 1: Fix react imports

examples/with-typescript/src/server.tsx
```diff
@@ -1,5 +1,5 @@
-import * as express from 'express';
-import * as React from 'react';
+import express from 'express';
```
jaredpalmer, line 1: turn on `esModuleInterop` flag

examples/with-typescript/tslint.json
```diff
@@ -26,7 +26,7 @@
     "no-arg": true,
     "no-bitwise": true,
     "no-consecutive-blank-lines": true,
-    "no-console": false,
+    "no-console": true,
```
jaredpalmer: This makes developing tough. We can keep this relaxed.

packages/razzle-plugin-typescript/index.js
```diff
@@ -1,21 +1,23 @@
 'use strict';
 
+const ForkTsCheckerWebpackPlugin = require('fork-ts-checker-webpack-plugin');
 const { babelLoaderFinder, eslintLoaderFinder } = require('./helpers');
 
 const defaultOptions = {
   useBabel: true,
```
jaredpalmer: Use babel should be false by default

jaredpalmer (changes requested):

examples/with-typescript/src/App.tsx
lucasterra: Why use `import * as React` when we have `esModuleInterop`?

jaredpalmer: You right sorry. I got confused for a sec

examples/with-typescript/src/client.tsx
jaredpalmer: ignore

packages/razzle-plugin-typescript/.eslintrc
```diff
@@ -0,0 +1,6 @@
+{
+  "env": {
+    "jest": true,
+    "node": true
+  }
+}
```
jaredpalmer, line 6: I think we should actually move eslint out of here. I think we ought to move it of the core as well, making a razzle-plugin-eslint.

packages/razzle-plugin-typescript/README.md
```diff
@@ -0,0 +1,56 @@
+# razzle-dev-utils
```
jaredpalmer: Change title to `Razzle Plugin TypeScript`

packages/razzle-plugin-typescript/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+This package contains a plugin for using TypeScript with Razzle
+
+## Usage in Razzle Projects
+
+```
+npm install --save razzle-plugin-typescript
```
jaredpalmer: use yarn

packages/razzle-plugin-typescript/README.md
```diff
@@ -0,0 +1,56 @@
...
+// razzle.config.js
+
+module.exports = {
+  plugins: [
+    'typescript'
+  ],
+};
+```
+
+Passing custom options to the plugin:
```
jaredpalmer: Make h2 called Options

packages/razzle-plugin-typescript/helpers.js
```diff
@@ -0,0 +1,30 @@
+'use strict';
+
```
jaredpalmer: Let's move this whole file to `razzle-dev-utils/pluginHelpers` because we'll probably reuse this all over the place

packages/razzle-plugin-typescript/.eslintrc
lucasterra: Oh, this was just to make linting work on the code and test files. It won't be used by plugin users.

jaredpalmer (changes requested):

jaredpalmer:
This is looking sweeeet

packages/razzle-plugin-typescript/.eslintrc
lucasterra: But yeah. I agree with removing eslint from the default webpack config.

jaredpalmer: Ahh sorry I just looked at the filename

packages/razzle-plugin-typescript/helpers.js
lucasterra: Yeah. Good point. 👍

packages/razzle-plugin-typescript/README.md
lucasterra: Copy/paste really is evil. 😈

lucasterra:
All done, I guess.

suciuvlad:
Looking forward to the merge 👍

romainquellec:
Nice work ! 👍

jaredpalmer (approved):

Issue #528, opened 2018-03-08
RFC: process.env.RAZZLE_RUNTIME_XXXX


jaredpalmer:
## Current Behavior

People struggle with how Razzle deals with `.env` variables (i.e. stringifying them at _build_ time by webpack) just like create-react-app.

## Expected Behavior

Razzle should have a way to honor runtime variables so users can more easily deploy their apps to Now/Heroku/Azure etc.

## Suggested Solution

Make `PORT`, `HOST`, and any other env variables prefixed with `RAZZLE_RUNTIME_XXXXXXX` available during runtime.

### Before 

**Available at compile time (i.e. will be stringified by webpack)**
- `RAZZLE_XXXXXX`
- `PORT`
- `HOST`

### After

**Available at runtime**
- `PORT`
- `HOST`
- `RAZZLE_RUNTIME_XXXXXXX`

**Available at compile time (i.e. will be stringified by webpack)**
- `RAZZLE_XXXXXX`

## Discussion

Another alternative is to _only_ stringify variables prefixed with `RAZZLE_XXXX` like the [razzle-heroku plugin](https://github.com/xouabita/razzle-heroku/blob/master/index.js) does. This would also be backwards compatible too. On the one hand, this would make it easier to work with how Heroku names it's config environment variables (e.g. `MONGO_URI`). On the other hand, this would be _too_ easy to mess up by accident (i.e. reference a runtime variable within shared isomorphic code that is `undefined` on the client...exploding the application).

## Related

#527 #526 #514 #477 #356 #285

gregmartyn:
I personally wouldn't worry too much about process.env being unavailable on the client. We already have to know that 'window' is unavailable on the server; it makes sense to me that process.env would have the opposite behavior. (there's no node process to have an env for, and browsers don't expose their env vars that I'm aware of) I think of process.env as a place where server secrets can end up, so I'd rather it was unavailable on the client-side by default.

If it were up to me, process.env.RAZZLE_INLINED_XXXXXX variables would get compiled in during the build, (and available on the client as inline DefinePlugin'd strings) and anything else would only be available on the server-side.

NODE_ENV could be inlined too, as it's mostly used as a description of the build and not something variable like the environment that the code is running in. There are also some performance reasons to do so.

I do like the idea of PORT and HOST being variable at runtime. I may be running the same build artifacts in different environments.

ferdinandsalis:
[...]

PR #611, opened 2018-05-13
Change env stringified so it doesn't overwrite process.env

tgriesser:
I believe this is part of the confusion around env vars and #528

With the current code, webpack's `DefinePlugin` sees the `process.env` variable as predefined entirely, `'process.env': ...`. If the env value doesn't exist webpack then inlines the entire stringified value object in the generated code.

With this change, we define each `process.env.$VAR` individually, which means that unknown variables will be left alone at build time.

If you add this in server.js:

```js
console.log(process.env.ITEMS)
console.log(process.env.NODE_ENV)
```

Current behavior:
```js
console.log(Object({"NODE_ENV":"development","PORT":"3000","VERBOSE":false,"HOST":"localhost","RAZZLE_ASSETS_MANIFEST":"/home/razzle-example/build/assets.json","BUILD_TARGET":"server","PUBLIC_PATH":"/","RAZZLE_PUBLIC_DIR":"/home/razzle-example/public"}).ITEMS);
console.log("development")
```

With this change:
```js
console.log(process.env.ITEMS)
console.log("development")
```

### DIFF
--- a/packages/razzle/config/env.js
+++ b/packages/razzle/config/env.js
@@ -81,12 +81,10 @@ function getClientEnvironment(target, options) {
       }
     );
   // Stringify all values so we can feed into Webpack DefinePlugin
-  const stringified = {
-    'process.env': Object.keys(raw).reduce((env, key) => {
-      env[key] = JSON.stringify(raw[key]);
-      return env;
-    }, {}),
-  };
+  const stringified = Object.keys(raw).reduce((env, key) => {
+    env[`process.env.${key}`] = JSON.stringify(raw[key]);
+    return env;
+  }, {});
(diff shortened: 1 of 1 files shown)


jaredpalmer:
Awesome! This looks very very promising. I need to run some tests on our infrastructure to check this out.

jaredpalmer:
Testing this out right meow.

jaredpalmer:
LGTM

tgriesser:
Awesome! Thanks!

s123121:
Any idea when we can have this pull request as a stable release? I really need this now to configure CI

PR #640, opened 2018-06-06
Add useEslint option to razzle-plugin-typescript

tgriesser:
Useful if you wish to use babel (for plugin transforms), but don't want to use eslint.

### DIFF
--- a/packages/razzle-plugin-typescript/README.md
+++ b/packages/razzle-plugin-typescript/README.md
@@ -31,6 +31,7 @@ module.exports = {
       name: 'typescript',
       options: {
         useBabel: false,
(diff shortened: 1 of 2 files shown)


jaredpalmer:
default to `false` plz

packages/razzle-plugin-typescript/README.md
```diff
@@ -31,6 +31,7 @@ module.exports = {
       name: 'typescript',
       options: {
         useBabel: false,
+        useEslint: true, // ignored if `useBabel` is false
```
jaredpalmer, line 34: oh i see. ignore my last comment. this makes sense.

PR #643, opened 2018-06-08
Add example: with-afterjs-react-native-web-apollo

noeljackson:
This is an example with after.js, react-native-web, and apollo. The reason for creating this was to show how you can integrate react-native-web with apollo, using AppRegistry, instead of bypassing it on the client side, as most examples I've seen try to bypass AppRegistry.

On the server and client, we are creating simple components that wrap After.js in an ApolloProvider and then send it off to AppRegistry, that integrates after.js. 

RNW as of 0.7.0 doesn't need a .babelrc with Razzle's env: https://github.com/necolas/react-native-web/issues/961#issuecomment-390796795

Prior art: 
- https://github.com/brunorzn/after.js/tree/with-apollo-example/examples/with-apollo/
- https://github.com/jaredpalmer/razzle/tree/master/examples/with-react-native-web


noeljackson:
@jaredpalmer is this kind of thing helpful? I have a few more examples to add involving after.js as well.

PR #649, opened 2018-06-11
Update babel-eslint to 8.x ...

daveli:
Update `babel-eslint` to 8.x so that decorators will work using `babel-plugin-transform-decorators-legacy`. See: https://github.com/babel/babel-eslint/issues/275.

Other option would be to update the docs and add documentation for the fix by adding manually the following depedency: `yarn add babel-eslint` or `npm install babel-eslint` (should install `babel-eslint` 8.x).

### DIFF
--- a/packages/razzle/package.json
+++ b/packages/razzle/package.json
@@ -32,7 +32,7 @@
     "assets-webpack-plugin": "3.5.1",
     "autoprefixer": "7.1.1",
     "babel-core": "6.25.0",
-    "babel-eslint": "7.2.3",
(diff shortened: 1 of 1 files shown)


atahani:
I appreciate it.
I've got an error at compiling client phase. 
`"razzle": "2.0.4"`
```
Module build failed (from ./node_modules/eslint-loader/index.js):
TypeError: Cannot read property 'type' of undefined
```
It has been fixed by installing `babel-eslint@8.1.2`. 🆗

PR #656, opened 2018-06-14
Bump Jest to v23

benhamlin:
This is just another version bump, 22 has an annoying issue where it's unable to update some snapshots when launched with --watch (as razzle test does). See https://github.com/facebook/jest/pull/5696.


jamesbhobbs:
It (22) also has a low prio vulnerability in braces. So yes we should do this.

jaredpalmer:
Released in 2.2.x

PR #660, opened 2018-06-16
preserve the full --inspect= flag as passed to yarn start

gregmartyn:
NodeJS's inspect flag takes an optional =IP:PORT. Razzle is currently ignoring that and treating it as a simple boolean.

Docs: https://nodejs.org/api/cli.html#cli_inspect_host_port

This PR renames the INSPECT_ENABLED env var to INSPECT, and stores the flag as it was passed to yarn start. (and the same for INSPECT_BRK_ENABLED)


jaredpalmer:
image

PR #668, opened 2018-06-18
Add Glamorous example

lucasterra:
Example with [Glamorous](https://github.com/paypal/glamorous) as the CSS-in-JS solution.

See `src/server.js` and `src/client.js` to see what is different from the basic implementation.


lucasterra:
Closing this, since Glamorous was deprecated in favor or Emotion.

PR #672, opened 2018-06-19
Add with-pwa example

rhodee:
Example with [Workbox](https://github.com/GoogleChrome/workbox) and [WebpackPWAManifest](https://github.com/arthurbergmz/webpack-pwa-manifest) Webpack plugins.

Most of the action is in `razzle.config.js`. The file `src/registerServiceWorker.js` was added and `src/index.js` was modified from the basic implementation.


lucasterra:
Maybe [write a plugin](https://github.com/jaredpalmer/razzle#writing-plugins) for this instead?

examples/with-pwa/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "my-razzle-app",
```
jaredpalmer, line 2: change pkg name to `razzle-examples-with-pwa` for congruency

examples/with-pwa/razzle.config.js
```diff
@@ -0,0 +1,31 @@
...
+    theme_color: '#ffffff',
+    background_color: '#ffffff',
+    related_applications: [],
+    icons: []
+}
+
+const modify = modifyBuilder({ pwaConfig, manifestConfig })
+
+module.exports = {
+    plugins: [{ func: modify }]
```
jaredpalmer, line 30: I'd prefer this API to be more inline with how the TS plugin works where you can pass an options object to the plugin.

examples/with-pwa/src/server.js
```diff
@@ -0,0 +1,54 @@
...
+  .use(express.static(process.env.RAZZLE_PUBLIC_DIR))
+  .get('/*', (req, res) => {
+    const context = {};
+    const markup = renderToString(
+      <StaticRouter context={context} location={req.url}>
+        <App />
+      </StaticRouter>
+    );
+
+    console.log(process.env)
```
jaredpalmer, line 21: Remove console.log

jaredpalmer (changes requested):
We can move the razzle-plugin-with-pwa discussion elsewhere I guess, but after you change the package name, this looks great. Thanks!

PR #616, opened 2018-05-17
Add GraphQL example

ghost:

PR #673, opened 2018-06-19
add apollo-redux example

ghost:
See this https://github.com/jaredpalmer/razzle/pull/616 for details


jaredpalmer:
Can we also do an example without Redux?

PR #691, opened 2018-07-03
Add pretty node errors

jackjocross:
I've been developing with Razzle and Jest and thought it could be helpful to have Jest's runtime error formatting for Razzle server side errors during development.

## Current server side errors
screen shot 2018-07-02 at 9 09 37 pm

## Server side errors with Jest formatting
screen shot 2018-07-02 at 9 08 17 pm

A couple of questions I had while developing:

1. There is a slight issue with the server stack traces where they include `/build/webpack:` in the path and always point to the first column of the line.  Should fixing stack traces be a prerequisite to adding this logging?
2. Is `razzle-dev-utils` the best place for this code to live?  Or would it be more useful to pull it out into a separate package that can be used on other projects.
3. Express has a default error middleware that forwards server side errors to the client.  Since Jest formatting targets the terminal, it doesn't look that great.  Is there anything Razzle should do to make those more readable, or is using the default express error middleware not a common use case?
screen shot 2018-07-03 at 10 52 28 am

Let me know what you think!

### DIFF
--- a/packages/razzle-dev-utils/prettyNodeErrors.js
+++ b/packages/razzle-dev-utils/prettyNodeErrors.js
@@ -1,0 +1,35 @@
+const {
+  formatExecError,
+  separateMessageFromStack,
+} = require('jest-message-util');
+
+function pretty(error) {
+  return `\n${formatExecError(
+    error,
+    { rootDir: process.cwd() },
+    {},
+    undefined,
+    true
+  )}`;
+}
+
+function usePrettyErrors(transform) {
+  const { prepareStackTrace } = Error;
+
+  Error.prepareStackTrace = (error, trace) => {
+    const prepared = prepareStackTrace
+      ? separateMessageFromStack(prepareStackTrace(error, trace))
+      : error;
+    const transformed = transform ? transform(prepared) : prepared;
+
+    return pretty(transformed);
+  };
+}
+
+// Currently needed to fix sourcemap path
+const stackTransform = ({ stack = '', ...rest }) => ({
+  stack: stack.replace('/build/webpack:', ''),
+  ...rest,
+});
+
+usePrettyErrors(stackTransform);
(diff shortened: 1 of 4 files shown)


packages/razzle-dev-utils/prettyNodeErrors.js
```diff
@@ -0,0 +1,35 @@
...
+      : error;
+    const transformed = transform ? transform(prepared) : prepared;
+
+    return pretty(transformed);
+  };
+}
+
+// Currently needed to fix sourcemap path
+const stackTransform = ({ stack = '', ...rest }) => ({
+  stack: stack.replace('/build/webpack:', ''),
```
jaredpalmer, line 31: I wonder if this has to do with some sort of memory filesystem within webpack?

jaredpalmer: Ahh it looks like it does. It's called "sourcemap namespacing." I'm not 100% sure we should remove this. 

https://github.com/webpack/webpack/blob/fe0eb45079019281cb536ef887adf301222e5a63/test/configCases/source-map/namespace-source-path/index.js

@nathanforce's suggestion in Slack about copying CRA is a good idea i think. Checkout `formatWebpackErrors.js` in `react-dev-utils`

jaredpalmer: Here's where CRA pulls it out i think. https://github.com/facebook/create-react-app/blob/next/packages/react-dev-utils/formatWebpackMessages.js#L112

jaredpalmer: Let's add a comment in the source that references the line of CRA with

```js
// Clean up Webpack's sourcemap namespacing in error stacks
// @see https://github.com/facebook/create-react-app/blob/next/packages/react-dev-utils/formatWebpackMessages.js#L112
```

PR #695, opened 2018-07-10
fix loadable method call

destructobeam:
need to call preloadAll here to avoid rendering loading content on client hydrate


destructobeam:
Nope, this is incorrect, I must be doing something wrong my side.

PR #698, opened 2018-07-12
upgrade razzle in examples

ambar:


brody2consult:
> Holler if this is a mistake

**looks like a mistake**

Issue #595, opened 2018-04-29
webpack-dev-server-speedy


hornta:
What is this dependency?
Why not use webpack-dev-server?

jaredpalmer:
https://medium.com/@kenneth_chau/speeding-up-webpack-typescript-incremental-builds-by-7x-3912ba4c1d15

mattlubner:
Could we remove the dependency on `webpack-dev-server-speedy` now?

From the linked article:
> Update: the webpack-dev-server maintainers had accepted my patch! So, go ahead and use webpack-dev-server@3.1.2. Personally, I have observed a slight 0.5s regression between 2.x release and the 3.x release, so we’re keeping the 2.x for now until we can move to using webpack-serve.

(The latest version of `webpack-dev-server` is now v3.1.4, whereas `webpack-dev-server-speedy@^3.1.1` is currently used by `razzle`)

jaredpalmer:
500ms is non trivial. Let’s try it out for ourselves. Can you submit a PR?

mattlubner:
Agreed, 500ms is a lot. However, and I may have misunderstood, but I thought @kenotron meant they're using `webpack-dev-server-speedy@2.11.2`, which is (was?) 500ms slower than the version used by `razzle` (`webpack-dev-server-speedy@^3.1.1`). That'd mean `razzle` already incurred this 500ms performance hit.

I'll put together a PR! 🙂

PR #701, opened 2018-07-14
Update to webpack-dev-server@^3.1.4

mattlubner:
Just a simple PR to switch back to `webpack-dev-server` (from the forked `webpack-dev-server-speedy`).

See: #595


mattlubner:
What's a good, reproducible way to measure how this change affects compile time?

Also (just curious), are the e2e tests intentionally not run in CI?

jaredpalmer:
Let’s cut a branch experimental release with this. Then we’ll be able to assess. One of our internal apps is >200,000 loc.

mattlubner:
How do we cut an experimental release? Is that something I can do?

PR #723, opened 2018-08-15
Use the correct port when PORT is not set

bkonkle:
I'm running a custom config where `PORT` is not hardcoded into the bundle, because I'm using that variable in a Docker container and need it to be dynamic.

This solves usages of `webpackHotDevClient` when `PORT` is not set inside the bundle.

### DIFF
--- a/packages/razzle-dev-utils/webpackHotDevClient.js
+++ b/packages/razzle-dev-utils/webpackHotDevClient.js
@@ -27,7 +27,7 @@ ErrorOverlay.startReportingRuntimeErrors({
   launchEditorEndpoint: url.format({
     protocol: window.location.protocol,
     hostname: window.location.hostname,
-    port: parseInt(process.env.PORT, 10) + 1 || window.location.port,
+    port: parseInt(process.env.PORT || window.location.port, 10) + 1,
     pathname: launchEditorEndpoint,
   }),
   onError: function() {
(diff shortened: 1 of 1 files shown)


jaredpalmer:
I’m on my phone but will release this later today or tomorrow. 

Can you add some docs in another PR about this?

gregmartyn:
This was a dupe of #581.
#581 also handles the case when window.location.port is 80 or 443
I've resolved the conflicts this created in #581 so that that's mergeable

marcinvalas:
As internally  `process.env.PORT_DEV` is used [see this](https://github.com/jaredpalmer/razzle/blob/master/packages/razzle-dev-utils/setPorts.js#L13) then port value should not be calculated, instead `process.env.PORT_DEV` should be used.

Moreover I believe that variables `PORT` (and `PORT_DEV`) should be compiled into bundle only if `process.env.NODE_ENV === 'development'` as on production we would like to start app on diffrent port without rebuilding.

gregmartyn:
More on that subject here: https://github.com/jaredpalmer/razzle/pull/740#pullrequestreview-153589566

Agreed it'd be nice to use an env var. I'm not at a computer that can run Razzle right now, but IIRC the issue is that PORT_DEV isn't one of the vars that gets inlined, so it's not available on the client-side. I'd like to add that, but it looks like that'd be a separate PR + discussion.

#581 is just updating the existing code to handle the edge-case of window.location.port returning "" instead of 80/443 if using those ports.

PR #752, opened 2018-09-12
remove mention of extract text plugin from react native web example, closes #744

### LINKED ISSUE #744: Remove extract-text-webpack-plugin mention from react-native-web example
It seems like extract-text-webpack-plugin was removed from razzle. The react-native-web example still mentions it. 

https://github.com/jaredpalmer/razzle/blob/master/examples/with-react-native-web/README.md

MrLoh:


necolas:
Just browsing and saw this PR. Also noticed that the SSR HTML shell example for RNW needs a slight adjustment to apply styles (necessary for full-page apps) to the various root elements:

https://github.com/necolas/react-native-web/blob/master/docs/guides/server-side-rendering.md

MrLoh:
@necolas could you just create another PR, since this was already merged. The setup with react-native-web and razzle/after.js works quite nicely. But because I am using after.js I do not have the same server side setup as in the readme you linked.

PR #765, opened 2018-09-27
Open the browser when razzle is first started

EmilTholin:
It might be a nice addition to open the browser automatically when razzle is first started, like Create React App does.

### DIFF
--- a/packages/razzle/scripts/start.js
+++ b/packages/razzle/scripts/start.js
@@ -9,6 +9,7 @@ const createConfig = require('../config/createConfig');
 const devServer = require('webpack-dev-server-speedy');
 const printErrors = require('razzle-dev-utils/printErrors');
 const clearConsole = require('react-dev-utils/clearConsole');
+const openBrowser = require('react-dev-utils/openBrowser');
 const logger = require('razzle-dev-utils/logger');
 const setPorts = require('razzle-dev-utils/setPorts');
 
@@ -50,6 +51,7 @@ function main() {
   // Compile our assets with webpack
   const clientCompiler = compile(clientConfig);
   const serverCompiler = compile(serverConfig);
+  let isInWatchMode = false;
 
   // Start our server webpack instance in watch mode after assets compile
   clientCompiler.plugin('done', () => {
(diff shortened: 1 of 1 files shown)


packages/razzle/scripts/start.js
```diff
@@ -59,7 +61,13 @@ function main() {
         stats: 'none',
       },
       /* eslint-disable no-unused-vars */
-      stats => {}
+      stats => {
+        // Just open the browser when the watch mode is first activated
+        if (!isInWatchMode) {
+          openBrowser(`http://localhost:${process.env.PORT}`);
```
olehreznichenko, line 67: For example, my razzle project is on `https`. How to handle this case?

EmilTholin: `localhost` should probably be something like `process.env.HOST || options.host || 'localhost'` instead, but how are you using `https` in development? I didn't know razzle could do that.

olehreznichenko: https setup was easy.

1. OverridepublicPath for client and server
config.output.publicPath = 'https://localhost:3001/'
2. And for devServer add certificates
```
Object.assign(config.devServer, {
			https: {
				key: readFileSync('./path/localhost.key'),
				cert: readFileSync('./path/localhost.crt')
			}
		})
```
3. Also connect certificates to express or any other server.
Most problem generated for localhost this certificates. and make then trusted for a machine.

EmilTholin: Alright, that's neat. I was under the impression that setting `process.env.PUBLIC_PATH` or changing `webpack.config.output.publicPath` was only supported in the build script.

Will have to think about the protocol issue some more.

olehreznichenko: Maybe add new env PROTOCOL?

MioQuispe:
Could this be done as a plugin maybe?

EmilTholin:
Yeah, that's a good idea. Not optimal to force the browser opening on users.

joeferraro:
Was this ever implemented as a plugin?

PR #774, opened 2018-10-05
Added with-react-router-4 example

DPr00f:
Just a simple example of how to use razzle with React Router 4.

@jaredpalmer let me know if more info is needed.


vidann1 (approved):

nimaa77:
thanks for your contribution, can you update this to use react-router-5, or maybe wait until react-router-6 comes out 😎

examples/with-react-router-4/src/server.js
```diff
@@ -0,0 +1,59 @@
...
+  const ServerApp = (
+    <StaticRouter location={req.url} context={routerContext}>
+      <RootApp />
+    </StaticRouter>
+  );
+
+  const markup = renderToString(ServerApp);
+
+  if (routerContext.url) {
+    res.status(302).setHeader('Location', routerContext.url);
```
nimaa77, line 25: I think you should use `routerContext.statusCode` instead of `302` here
```suggestion
    res.redirect(routerContext.statusCode, routerContext.url);
```

examples/with-react-router-4/src/server.js
```diff
@@ -0,0 +1,59 @@
...
+    res.end();
+    return;
+  }
+
+  res.status(routerContext.missed ? 404 : routerContext.serverStatusCode || 200)
```
nimaa77, line 30:
```suggestion
  res.status(staticContext.statusCode || 200)
```

nimaa77 (changes requested):

nimaa77:
closed in favor of #1278

PR #841, opened 2018-11-15
Add GraphQL file loader plugin

Evanion:
This is a first pass at a proposal for Graphql AST file support.
It's a plugin that adds support for `*.graphql` and `*.gql` AST files used to store Graphql queries in your codebase.

Known issues:
* Tests - graphql-tag loader needs to be added as a string. But `makeFinderLoader` doesn't seem to support that.
* Jest config needs to have a [transformer](https://github.com/remind101/jest-transform-graphql) in order to add graphQL "support".


fivethreeo:
added in finch branch

PR #877, opened 2019-01-04
Fix startReportingRuntimeErrors warning

oyvinmar:
Fixes the startReportingRuntimeErrors warning reported in https://github.com/jaredpalmer/razzle/pull/839.

Mainly copied from https://github.com/facebook/create-react-app/commit/00ed100b26adc519fd90e09ebffd83c8d7dc4343#diff-5166b8ed2c51bdab6a247a95b14ff595 but with changes to open on the correct port.

Needed to add a `{ mode: 'no-cors' }` to prevent a CORS issue. I have not explored if there is a better workaround.


jaredpalmer:
Cutting a new alpha release now

PR #883, opened 2019-01-13
Add with-redux-observable example

christophediprima:
Add an example using following:

- Typescript
- Redux-Observable
- Found router


brody2consult:
??

christophediprima:
@brodybits there is a lot of pending MR if that's why you are asking?? I don't really know what is the reason they don't get merged... I guess the managers are missing time to merge them...

brody2consult:
Hi @christophediprima I entered the "??" comment to stop the stale bot from closing this PR.

The good part is that they *did* make a recent 3.0 release, merged a few PRs including #967 from myself.

In general I think this project has progressed to a point where it needs more active management and support. I think that @jaredpalmer should either add some more active members or consider submitting to an organization such as [Open JS Foundation](https://openjsf.org/).

PR #885, opened 2019-01-14
[v3] Client-only mode

jaredpalmer:
This adds a client-only mode to Razzle so that you can build SPA's in addition to isomorphic web apps. There are still some more optimizations we can shamelessly steal from CRA v2, but this at least gets us started. For now, you pass `--type=spa` to `razzle start` and `razzle build` and add a HTML template in `public/index.html` (just like CRA).  Note: I kept the client entry point as `client.js`, although we may want to switch it to `index.js`.

Future todos:

- Add a message to the logger stating that the SPA app is running on port 3000 during dev
- Verify that `public/index.html` exists
- Add instructions for adding or removing SSR


thecodedrift:
Just $0.02. I'd keep it as client.js. This way, if people are using razzle and find they need the isomorphic parts of razzle, it's straightforward to add an `index.js` and `server.js`

jaredpalmer:
Word. I agree

cesargdm:
@jaredpalmer Is it ready? 😬

examples/basic-spa/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "razzle-examples-basic-spa",
+  "version": "3.0.0-alpha.2",
+  "license": "MIT",
+  "scripts": {
+    "start": "razzle start",
```
thecodedrift: Need `--type=spa` here, correct?

packages/razzle/bin/razzle.js
```diff
@@ -1,40 +1,66 @@
 #!/usr/bin/env node
 'use strict';
 
+const sade = require('sade');
```
thecodedrift, line 4: Do we want to bundle the `sade` changes w/ the spa support? It wouldn't be a breaking change, but it does increase the change's surface area by quite a bit.

packages/razzle/scripts/start.js
```diff
@@ -60,24 +74,29 @@ function main() {
...
+        stats => {}
+      );
+    }
   });
 
   // Create a new instance of Webpack-dev-server for our client assets.
   // This will actually run on a different port than the users app.
   const clientDevServer = new devServer(clientCompiler, clientConfig.devServer);
-
+  const maybePlusOne = clientOnly ? 0 : 1;
```
thecodedrift: nit: should we just call this `portOffset`?

thecodedrift (commented):
Just a few minor things when I read through the PR. My biggest concern is making the bigger CLI changes at the same time (adding `sade` and `mri`), but if it makes it easier to add yet-another-option, then 👍

cesargdm:
@jaredpalmer Sorry for the insistence, but I would really love to see this option to see razzle as a more customisable option to create-react-app; thanks in advance

rvanlaarhoven:
@elliottjro This feature would be pretty useful in my case at the moment, since I'm about to setup a client-only app and would like to stick with Razzle if possible. I'll also have a look at the current state of this pull request to see if I can already use it.

elliottjro:
@rvanlaarhoven will take a look, been quite slow because I became quite sick, getting better now and just catching up on my dayjob before I tackle PRs again

rvanlaarhoven:
No problem, I already went ahead and resolved the conflicts on my fork (See https://github.com/rvanlaarhoven/razzle/tree/feature/client-only-mode), but haven't had the time yet to actually test anything.

nimaa77:
hello everyone, this awsome I did some changes, so this is our new todo list.

TODO:
- [x] Add a message to the logger stating that the SPA app is running on port 3000 during dev
- [x] Verify that public/index.html exists
- [x] Add instructions for adding or removing SSR
- [x] add some tests!

and there are two problems here:
1) there is an error in the console about socket and HMR, I wasn't able to fix it. (but HMR works 🤕)
this the error message `:3001/sockjs-node/info?t=1584140341987:1 Failed to load resource: net::ERR_CONNECTION_REFUSED`
2) when I changed the content of index.html file HMR or webpack crashed ...

### UPDATE 1
I fixed the problem of HMR ...

nimaa77:
this is ready to get merged 😎

please do a code review on this and then let's merge this

fivethreeo:
Nothing that jumps out at me on this. Seems sound. But it affects many parts so, if the tests don’t break it, get it in there and do a alpha/beta to get some real world feedback.

PR #888, opened 2019-01-16
[v3] Create razzle-plugin-eslint

bartlangelaan:
I removed everything that contained 'eslint' and moved it to the 'razzle-plugin-eslint' plugin.

One exception: the eslintFormatter is still in react-dev-utils. Should that move too? The react-dev-utils package still seems like a logic place, because that depends on all the logging packages.

Also, the errors and warnings don't show up in the console. They only show up when you visit the page. I'm not sure where the display of the console is handled at the moment, so I could use some help there! 😄


jaredpalmer:
Yeah I would just have razzle-plugin-eslint depend on react-dev-utils then

jaredpalmer:
Oh yeah, shoot I think it is handled inside of Razzle core. Let me look

jaredpalmer:
It's here. 

https://github.com/jaredpalmer/razzle/blob/8fed32b57ec5b75c421e04f00a4081da687ab6d6/packages/razzle/config/createConfig.js#L61

Just make `react-dev-utils` a dependency of the eslint plugin.

jaredpalmer:
The only bit we may put back one day is eslint in the ts plugin. ESLint just announced an initiative to support TS, but whatevs. This is more unix-like anyways.

mschipperheyn:
Shouldn't the eslint packages.json version not be in line with 3.3.0? It's currently set to 2.2.0

PR #891, opened 2019-01-19
[v2] Added with-apollo-typescript example 

zifahm:
Added with-apollo-typescript example.


fivethreeo:
Could you do this against the canary branch and use the builtin typescript support?

PR #899, opened 2019-01-27
With sequelize example

mschipperheyn:
With-sequelize example.

Highlights:
- Creates docker mysql container
- Creates database
- Seeds database
- Displays database contents on homepage

I can move this example to master if you prefer.


examples/with-sequelize/README.md
```diff
@@ -0,0 +1,51 @@
...
+Download the example [or clone the whole project](https://github.com/jaredpalmer/razzle.git):
+
+```bash
+curl https://codeload.github.com/jaredpalmer/razzle/tar.gz/master | tar -xz --strip=2 razzle-master/examples/with-sequelize
+cd with-sequelize
+```
+
+### Database server
+
+The example comes with a `docker` container to bootstrap a database server. For this to work you should have `docker` installed. If you already have `mysql` installed and want to use that server instead of `docker`, create a `.env` file in the `with-sequelize` root, with the following values:
```
jaredpalmer: I like this. I think we should rework it to assume that you connect to sqllite with just dotenv vars at first. It’s a sequelize example after all. then at the bottom let’s document how to spin up local docker mysql and and postgres as well as just installing these with brew?

jaredpalmer: Actually screw sqllite, I’m fine with it as mysql or postgres  as long as we can use a locally installed dB in the example (without docker)

examples/with-sequelize/src/Home.js
```diff
@@ -0,0 +1,64 @@
+import React from 'react';
+import logo from './react.svg';
+import './Home.css';
+
+class Home extends React.Component {
+  state = {
+    users: null,
+  };
+
+  async componentDidMount() {
```
jaredpalmer: This isn’t a good thing to promote. Refactor to just use promises. Make sure to include a polyfill for isomorphic fetch.

examples/with-sequelize/src/server.js
```diff
@@ -0,0 +1,81 @@
...
+import express from 'express';
+import { renderToString } from 'react-dom/server';
+import models, { sequelize, testConnection, close } from './sequelize';
+
+const assets = require(process.env.RAZZLE_ASSETS_MANIFEST);
+
+const app = express();
+const server = http.createServer(app);
+
+app.get('/rest/users', async (req, res) => {
```
jaredpalmer, line 14: Add a comment that this is a rest API route and not rendering react

examples/with-sequelize/src/sequelize/index.js
```diff
@@ -0,0 +1,33 @@
...
+
+Object.keys(models).forEach(key => {
+  const model = models[key];
+  if (model.associate) {
+    model.associate(models);
+  }
+});
+
+sequelize.addHook('afterBulkSync', async () => {
+  if (process.env.NODE_ENV !== 'test') await seed(models, sequelize);
```
jaredpalmer: Add curly braces

jaredpalmer (changes requested):
This is epic! You rock! Some little nits here and there.

mschipperheyn:
@jaredpalmer I have a question. I wonder if I should not perhaps add a test to promote api testing. The only thing is that I found that `globalSetup` and `globalTeardown` are not supported. Which is what I use to start a database server and force the creation of the table structure one time only. I test serially against a single database to prevent issues with various test creating and removing data and creating weird errors (I prefer integration testing against mock database testing).

I know some people like to use some wizardry to create a new database process with various ports to keep the parallel testing in place. I never tried it.

elliottjro:
@mschipperheyn for db setup, I usually put it into the package.json scripts as a setup and just add it as a requirement before running jest:

```json
"scripts": {
  "unit": "node ./jest-setup/migrate-db.js && node ./jest-setup/seed-db.js && npm run test"
}
```

where `migrate-db.js`
```js
const { spawn } = require('child-process-promise')

const run = async () => {
  const url = process.env.TEST_DB_URL || 'mysql://root@localhost:3306/database-test'
  try {
    await spawn('./node_modules/.bin/sequelize', ['db:migrate', `--url=${url}`], { stdio: 'inherit' })
    console.log('Migration success')
  } catch (e) {
    console.error('migration failed')
    process.exit(1)
  }
  process.exit(0)
}
run()
```

If you like, you can modify this PR (I know its quite old now, sorry for the delay) or I can just merge it in as is and you can make another PR when you have time to add this in?

elliottjro (approved):
Looks like all review change requests has been addressed

nimaa77:
@elliottjro hello, why you don't merge this 😎?

fivethreeo:
Could you do this against the finch branch?

fivethreeo:
Um, finch branch is now canary, can you do this against canary?

PR #903, opened 2019-01-29
Add globalSetup, globalTeardown and moduleDirectories as jest options

mschipperheyn:
This change liberates three additional options that are imho crucial for testing server side development:
- globalSetup: allow you to prep, and seed a database, start a server
- globalTeardown: allow you to stop a database, stop a server, gracefully shutdown
- moduleDirectories: The jest complement to `.babelrc` => `module-resolver`

`module-resolver` allows you to `absolute` import components that are not in `node_modules`, e.g. `import TextInput from 'components/form/TextInput`, where `components` is placed in `src/components`. It helps keep your code more readable and not have endless `../../../components` style imports.

When you use this, jest also needs to know, and `moduleDirectories` is the key for it.

I didn't quite understand why these keys were blocked, but perhaps this is a relic of `create-react-app` that for obvious reasons doesn't bear the burden of server side testing.


mschipperheyn:
I can see snyk is failing, but I don't have the right access level to be able to diagnose why

PR #907, opened 2019-02-01
console log custom razzle config

mschipperheyn:
Small change to log the use of a custom `razzle.config.js`. It's always nice to know when things don't work as expected that the custom razzle config was actually loaded.

What lead me to this PR was a typo in my `razzle.config.js` that created an unlogged error, bc the try/catch doesn't log anything if something goes wrong in processing the razzle.config.js. With this console.log, you at least know what's happening (or going wrong)


omdjin (approved):

packages/razzle/scripts/build.js
```diff
@@ -74,8 +74,12 @@ function build(previousFileSizes) {
   let razzle = {};
   try {
     razzle = require(paths.appRazzleConfig);
+    console.log(chalk.yellow('Found custom razzle.config.js'));
```
jaredpalmer: Yellow should be reserved for warnings.
```suggestion
    console.log(chalk.dim('Found custom razzle.config.js'));
```

packages/razzle/scripts/build.js
```diff
@@ -74,8 +74,12 @@ function build(previousFileSizes) {
...
     /* eslint-disable no-empty */
-  } catch (e) {}
+  } catch (e) {
+    const errMessage = chalk.yellow(e.message);
+    console.log(`No custom razzle config processed (${errMessage})`);
```
jaredpalmer: The problem with this though is that this will _always_ show if there is no `razzle.config.js` (ie. the require() fails).

jaredpalmer (changes requested):

nimaa77:
closed in favor of #1231

PR #930, opened 2019-02-20
Add target and env to modifyBabelOptions

revskill10:


aga5tya:
@jaredpalmer Could you please review this, would help in my use-case too where I have set of babel plugins that is needed only on the server compilation.

packages/razzle/config/createConfig.js
```diff
@@ -54,6 +54,8 @@ module.exports = (
     babelrc: true,
     cacheDirectory: true,
     presets: [],
+    target,
+    env,
```
aga5tya: We could move this as an additional argument to modifyBabelOptions ?

modifyBabelOptions(mainBabelOptions, { target, env })

revskill10: I moved them to a `buildOptions`.

phelma:
This would be really useful to have seperate babel configs for browser and node, specifically seperate @babel/preset-env browserslist targets.

nimaa77:
Hello and thanks for your contribution we have the same config in #1108 and it's merged.

so I'm going to close this at this moment.

PR #941, opened 2019-03-02
make `paths.appPublic` optional

ambar:


nimaa77:
@ambar hello, can you explain why this is useful or link it to any issue?

fivethreeo:
fixed in dev

Issue #900, opened 2019-01-27
Zeit Now 2.0 example?


ukyiwin:

jaredpalmer:
Feel free to open a PR

Siilwyn:
Think this would replace the current Now example since 1.0 seems to be on its way out.

chrisdrackett:
I came here to open then as I am getting a build error currently when using `@now/node-server`. Has anyone else had luck getting razzle to build on now?

revskill10:
@chrisdrackett It's not hard .
BUT in my case, i have to "prebuild" the app with manual `ncc` , instead of using `now` as CI.

chrisdrackett:
yeah, I can get it going with a pre-build, but I'd much rather build on deploy if possible

lessp:
hmm, I can't figure out a reasonable way to pre-build it. @revskill10, you said something about `ncc`, would you mind elaborating?

bausk:
@lessp `ncc` is Zeit's transpiler for Node.js that bundles dependencies into a single file, it can be run locally as an npm script and, to my understanding, works similarly to Zeit's builders like `@now/node` and `@now/node-server`.

I have just tried using it to deploy Razzle and found it to be just repetition of the standard Zeit build process, at least for the standard Razzle template with no dependencies. Here's my `now.json`, for reference:

```
{
    "version": 2,
    "name": "example",
    "builds": [
        {
            "src": "build/public/**",
            "use": "@now/static"
        },
        {
            "src": "build/server.js",
            "use": "@now/node-server"
        }
    ],
    "routes": [
        { "src": "/assets.json", "dest": "build/assets.json" },
        { "src": "/favicon.ico", "dest": "build/public/favicon.ico" },
        { "src": "/robots.txt", "dest": "build/public/robots.txt" },
        { "src": "/static/(.*)", "dest": "build/public/static/$1" },
        { "src": "/(.*)", "dest": "build/server.js" }
        
    ]
}
```
Maybe someone knows how to get rid of these abominations of routes BTW.

PR #942, opened 2019-03-02
Added example with-now-v2

dfreire:
Straightforward example to deploy to now v2.

Also see #900


examples/with-now-v2/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "razzle-examples-now-v2",
+  "version": "2.4.1",
+  "license": "MIT",
+  "scripts": {
+    "start": "razzle start",
+    "build": "razzle build",
+    "test": "razzle test --env=jsdom",
+    "start:prod": "NODE_ENV=production node build/server.js"
```
sync: You could also add a now-build script

examples/with-now-v2/package.json
```diff
@@ -0,0 +1,21 @@
...
+  },
+  "dependencies": {
+    "express": "^4.16.4",
+    "razzle": "^2.4.1",
```
jaredpalmer: Duplicate deps. Only include devDep

examples/with-now-v2/package.json
```diff
@@ -0,0 +1,21 @@
...
+    "react": "^16.8.3",
+    "react-dom": "^16.8.3",
+    "react-router-dom": "^4.3.1"
```
jaredpalmer: Remove react router since it isn’t used

jaredpalmer (changes requested):
Few nits

gl0gl0:
I've tried this, but I'm getting an error after now deployment that says:
`INTERNAL_MISSING_RESPONSE_FROM_CACHE`
Any ideas? I can't find any documentation about it.

PR #945, opened 2019-03-03
refactor (with-material-ui) upgrades to @material-ui/core

potench:
## Background

 - [x] Upgrade material-ui example to use @material-ui/core 
 - [x] Add jest as dependency to material-ui example (fixes error during `yarn test`)
 - [x] Add react-jss as dependency (fixes error during `yarn start`)

### Fixes the Following Errors
```
// During `yarn test`
TypeError: environment.teardown is not a function

// During `yarn start` 
razzle/examples/with-material-ui/build/server.js:41241
  var sheetsRegistry = new !(function webpackMissingModule() { var e = new Error("Cannot find module \"react-jss\""); e.code = 'MODULE_NOT_FOUND'; throw e; }())();
```

## Test Plan
```
cd razzle
yarn
cd examples/with-material-ui
yarn test # should pass
yarn start # browse to http://localhost:3000/ and make sure it loads 
```

### DIFF
--- a/examples/with-material-ui/src/server.js
+++ b/examples/with-material-ui/src/server.js
@@ -3,9 +3,8 @@ import React from 'react';
 import express from 'express';
 import theme from './theme';
 import jss from './styles';
-import { SheetsRegistry } from 'react-jss';
-import { JssProvider } from 'react-jss';
-import { MuiThemeProvider } from 'material-ui/styles';
+import { JssProvider, SheetsRegistry } from 'react-jss';
+import MuiThemeProvider from '@material-ui/core/styles/MuiThemeProvider';
 import { renderToString } from 'react-dom/server';
 
 const assets = require(process.env.RAZZLE_ASSETS_MANIFEST);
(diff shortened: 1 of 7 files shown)


elliottjro:
@potench Sorry for the late reply, could you please fix the merge conflicts and I will merge in soon as you are done. Thanks again for doing this!

examples/with-material-ui/src/server.js
```diff
@@ -3,9 +3,8 @@ import React from 'react';
 import express from 'express';
 import theme from './theme';
 import jss from './styles';
-import { SheetsRegistry } from 'react-jss';
-import { JssProvider } from 'react-jss';
-import { MuiThemeProvider } from 'material-ui/styles';
+import { JssProvider, SheetsRegistry } from 'react-jss';
```
sibelius, line 6: there is no need for jssprovider anymore, check the new material ui ssr api

nimaa77:
closed in favor of #1278

PR #948, opened 2019-03-07
fix: remove invalid option for `MiniCssExtractPlugin`

ambar:
It's an option from `extract-text-webpack-plugin`

https://github.com/webpack-contrib/mini-css-extract-plugin/search?q=allChunks


s123121:
this option helps prevent FOUC in production for razzle back in the day, after changing to `mini-css-extract-plugin`, the main css bundle does not includes all the styles

nimaa77 (approved):

nimaa77:
thanks ❤️

Issue #509, opened 2018-02-12
Add Apollo Client SSR example


michaelmerrill:
https://www.apollographql.com/docs/react/recipes/server-side-rendering.html

romainquellec:
Is there something specific to know about Apollo and after.js ?

rizalibnu:
@michaelmerrill  check this out for example [#950](https://github.com/jaredpalmer/razzle/pull/950)

PR #950, opened 2019-03-07
With Apollo GraphQL Example

rizalibnu:
Add an example of Razzle with Apollo GraphQL as requested on [#509](https://github.com/jaredpalmer/razzle/issues/509)

This sample has the full SSR setup as described in the [apollo graphql docs](https://www.apollographql.com/docs/react/features/server-side-rendering.html).


zifahm:
PR for apollo with typescript example #891 
also readme has been updated.

examples/with-apollo-graphql/src/server.js
```diff
@@ -0,0 +1,88 @@
...
+  .use(express.static(process.env.RAZZLE_PUBLIC_DIR))
+  .get('/*', (req, res) => {
+    const Application = (
+      <ApolloProvider client={client}>
+        <App />
+      </ApolloProvider>
+    );
+
+    // during request
+    getDataFromTree(Application).then(() => {
```
revskill10, line 73: Should we catch error here ?

rizalibnu: I think we should prevent getDataFromTree errors and handle them in components via data.error
[https://www.apollographql.com/docs/react/basics/queries.html#graphql-query-data-error](https://www.apollographql.com/docs/react/basics/queries.html#graphql-query-data-error)

examples/with-apollo-graphql/src/client.js
```diff
@@ -0,0 +1,28 @@
...
+
+import App from './App';
+
+// Create an ApolloClient instance
+// Pass your GraphQL endpoint to uri
+const client = new ApolloClient({
+  link: createHttpLink({
+    uri: 'https://gql-placeholder.herokuapp.com/graphql',
+  }),
+  cache: new InMemoryCache(),
```
rozenmd, line 16: Working from the Apollo docs (https://www.apollographql.com/docs/react/features/server-side-rendering#store-rehydration), you'll want to swap this out for
```
cache: new InMemoryCache().restore(window.__APOLLO_STATE__)
```
might also want to add 
```
ssrForceFetchDelay: 100,
defaultOptions: {
  query: {
    fetchPolicy: 'cache-and-network',
  },
},
``` 
to avoid fetching data on the client-side that the server has already fetched.

rozenmd (changes requested):

examples/with-apollo-graphql/src/client.js
```diff
@@ -0,0 +1,28 @@
...
+import { createHttpLink } from 'apollo-link-http';
+import { InMemoryCache } from 'apollo-cache-inmemory';
+
+import App from './App';
+
+// Create an ApolloClient instance
+// Pass your GraphQL endpoint to uri
+const client = new ApolloClient({
+  link: createHttpLink({
+    uri: 'https://gql-placeholder.herokuapp.com/graphql',
```
stevengrimaldo, line 14: This URL doesn't appear to be valid any longer when directly accessed. The following URL does work though: http://gql-placeholder.herokuapp.com/playground.

PR #953, opened 2019-03-11
Add moduleDirectories jest parameter

thepratt:
I had the use-case of including compiled purescript output (in the `output` dir) in order to run tests. Meaning I had the the need to set moduleDirectories to `["node_modules", "output"]` in order to use razzle commands.

Support moduleDirectories jest configuration parameter for including compiled-to-js output in test run.


nimaa77 (approved):

nimaa77:
yes, why not :)

PR #967, opened 2019-04-03
remove extra Razzle filter from with-react-native-web

brody2consult:
Example seems to work without it if I would use Yarn (there seems to be an issue with using `npm start`).


brody2consult:
Thanks for merging

PR #1020, opened 2019-05-20
update hmr-for-port-80 to razzle v3

gregmartyn:
Store the client port in a var instead of repeatedly doing an inline calculation

Handle window.location.port = '', which happens if port matches the protocol (80 for http, 443 for https)


fivethreeo:
https://github.com/jaredpalmer/razzle/pull/1143

A more comprehensive solution.

Issue #1019, opened 2019-05-20
add support for programmatic usage and client only builds (i.e. for electron target)


alidcast:
ref #392 and #320 (issues requesting electron support, shut down because razzle runs two bundles by default)

seems like there are two issues with using razzle with electron 
1) there's no programmatic usage 
2) there's no option to only run client bundle 

would you be interested in a PR adding these two options? it would align with other SSR built tools like next.js and nuxt.js which also offer the above and thus support electron as a target

I currently have a fork that has a working version of this. it extracts some of the logic inside `scripts` so that the client bundle can be compiled programmatically, and has a minor change in `createConfig` to fix an error that was occurring in electron window.

PR #1025, opened 2019-06-03
expose config logic and generalize web target to support electron

alidcast:
works on #1019

since webpack supports electron as a target, would be great if razzle added support for it as well

this PR exposes the config logic, so we can create custom start script for electron that only runs client bundle (the only issue I ran into was that `multiStep` option was causing errors, which is why I generalized the target client env to fix this, so now we have `IS_CLIENT` and `IS_ELECTRON` checks)

a custom electron `script` can also be added in another PR to make setup even easier, or it can be shown as an `example`

### DIFF
--- a/packages/razzle/config/createConfig.js
+++ b/packages/razzle/config/createConfig.js
@@ -68,7 +68,8 @@ module.exports = (
 
   // Define some useful shorthands.
   const IS_NODE = target === 'node';
-  const IS_WEB = target === 'web';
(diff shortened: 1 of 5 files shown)


madisvain:
This seems like quite an elegant solution. Is there a reason why Razzle does not want to support Electron as a use case?

alidcast:
not sure but this is an old PR so going to go ahead and close

PR #1033, opened 2019-06-18
use onlyLocals option instead of css-loader/locals

dominicrico:
`css-loader/locals` is no longer available in favor of the new `onlyLocals` options.

### DIFF
--- a/packages/razzle-plugin-scss/index.js
+++ b/packages/razzle-plugin-scss/index.js
@@ -103,8 +103,8 @@ module.exports = (
       use: isServer
         ? [
             {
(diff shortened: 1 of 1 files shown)


wickedev:
I still get an error. I think we'll need a merge for fix this. "Cannot find module 'css-loader/locals'"

pbassut:
Works for me. @jaredpalmer can we get this in?

jtsui:
This PR works for me thanks for creating @dominicrico and merging @jaredpalmer .

Are there plans to release a new version of `razzle-plugin-scss` that includes this change? Looks like there have been several changes and no new releases of the plugin.

Issue #1035, opened 2019-06-20
Production build hangs


vuhrmeister:
The build hangs, it never stops. That's the output:

```shell
> razzle build

Creating an optimized production build...
Compiling client...
Compiled client successfully.
Compiling server...
Compiled server successfully.
Compiled successfully.

File sizes after gzip:

  65.76 KB  build/static/js/bundle.8241e039.js
```

After that nothing happens anymore, I need to cancel the process.

Obviously the build succeeded but it does not actually quit the process.

I'm on version 3.0.0 with no custom configuration or plugins.

vuhrmeister:
I just tried to locate the issue. Obviously it's an issue with tree shaking. I'm using Material-UI and import my components like so:

```javascript
import { CssBaseline } from '@material-ui/core'
```

But if I import them by path, then the build runs just fine:

```javascript
import CssBaseline from '@material-ui/core/CssBaseline'
````

Unfortunately there is no error output.

Running the app in development mode works just fine, no matter the import style!

**Comment by vuhrmeister** [comment]
Also there are circumstances where the build output stops even earlier:

```shell
> razzle build

Creating an optimized production build...
Compiling client...
Compiled client successfully.
Compiling server...
```

I have a component like:

```javascript
const Footer = () => (
  <footer>
    <Component1 />
    <Component2 />
  </footer>
)
```

If I either remove `<Component1 />` or `<Component2 />` the process runs further until hanging after displaying the file size (like described above).

vuhrmeister:
I checked out the Material-UI example and fixed it so that it's working again. Here's the PR: https://github.com/jaredpalmer/razzle/pull/1041

And here is a version that breaks the build as described above: https://github.com/vuhrmeister/razzle/compare/fix-material-ui-example...material-ui-broken-build
I only changed the import statement.

Unfortunately this is the easiest case. In my (larger) application it breaks builds also with the (in the example) working import. There must be something else, but probably related to this one.

gugaevkirill:
I have the same issue and I've noticed, that adding semicolon after import fixes the issue:
`import { CssBaseline } from '@material-ui/core'`

gugaevkirill:
I've got the same Issue:
This import hangs production build:
```
import {ConnectedRouter} from 'connected-react-router';
```

But if I add another unused import after it, it will work:
```
import {ConnectedRouter} from 'connected-react-router';
import BrowserRouter from 'react-router-dom/BrowserRouter';
```

[...]

PR #1041, opened 2019-06-27
Fix Material-UI example

vuhrmeister:
This example was using a really old version of Material-UI and broke the build. I updated it to use the latest Material-UI version.


examples/with-material-ui/src/server.js
```diff
@@ -2,10 +2,8 @@ import App from './App';
 import React from 'react';
 import express from 'express';
 import theme from './theme';
-import jss from './styles';
-import { SheetsRegistry } from 'react-jss';
-import { JssProvider } from 'react-jss';
-import { MuiThemeProvider } from 'material-ui/styles';
+import { MuiThemeProvider } from '@material-ui/core/styles';
```
irontitan76: Use `import { ServerStyleSheets, ThemeProvider } from '@material-ui/styles';`.

examples/with-material-ui/src/server.js
```diff
@@ -16,18 +14,15 @@ server
...
+      sheets.collect(
+        <MuiThemeProvider theme={theme}>
           <App />
         </MuiThemeProvider>
-      </JssProvider>
+      )
     );
-    const css = sheetsRegistry.toString();
+    const css = sheets.toString();
     res.send(
```
irontitan76: You may want to explicitly send a 200. `res.status(200).send(...)`. :)

examples/with-material-ui/src/App.js
```diff
@@ -1,31 +1,30 @@
 import './App.css';
 import React from 'react';
 import PropTypes from 'prop-types';
-import { withStyles } from 'material-ui/styles';
-import Paper from 'material-ui/Paper';
-import Typography from 'material-ui/Typography';
+import Paper from '@material-ui/core/Paper';
```
irontitan76, line 4: Since Material-UI now does tree-shaking, these imports can be safely changed to `import { Paper, Typography } from '@material-ui/core'`. That being said, webpack >= 2.0 or parcel with a certain flag is needed. https://material-ui.com/guides/minimizing-bundle-size/#how-to-reduce-the-bundle-size

examples/with-material-ui/src/client.js
```diff
@@ -1,14 +1,11 @@
...
-import { MuiThemeProvider } from 'material-ui/styles';
+import { MuiThemeProvider } from '@material-ui/core/styles';
 import App from './App';
 
-// This is needed in order to deduplicate the injection of CSS in the page.
-const sheetsManager = new WeakMap();
-
 hydrate(
-  <MuiThemeProvider sheetsManager={sheetsManager} theme={theme}>
+  <MuiThemeProvider theme={theme}>
```
irontitan76: ThemeProvider

irontitan76 (changes requested):
You beat me to it! I left a few comments, but otherwise it looks good.

examples/with-material-ui/src/App.js
vuhrmeister: That's the point, this would break razzle as mentioned in this issue: https://github.com/jaredpalmer/razzle/issues/1035 
And here the linked change: https://github.com/vuhrmeister/razzle/compare/fix-material-ui-example...material-ui-broken-build

vuhrmeister:
Thx @irontitan76 
You're absolutely right with the theme provider. Will change that.

examples/with-material-ui/src/server.js
vuhrmeister: hmm … honestly I wouldn't. But since the official create-razzle-app is doing so, I can change it.

examples/with-material-ui/src/App.js
irontitan76: Ah, interesting. I'll have to take a look at that one, because I haven't come across that issue in any of my environments yet. But I agree, leave it as is!

gobengo:
@vuhrmeister Thank you for contributing this. I found it useful in building [razzle-examples-with-material-ui-typescript](https://github.com/gobengo/razzle-examples-with-material-ui-typescript), which is what you did + typescript. (Who knows what the naming convention should be for examples with more than one thing going on?). <3

PR #1047, opened 2019-07-06
added cssModules for sass, css in  createConfig.js

astranavt:
Added support for css modules for *.module.css and *.module.scss files in webpack configuration

The following libraries have been updated:
"css-loader": "^3.0.0",
 "mini-css-extract-plugin": "^0.7.0",
"postcss-preset-env": "^6.6.0",
 "react-dev-utils": "^9.0.1",
"webpack": "^4.35.2",

The following libraries have been added:
 "node-sass": "^4.12.0",
 "postcss-normalize": "^8.0.1",
"sass-loader": "^7.1.0",


fivethreeo:
I think this should be two separate pulls, one for .module.css for razzle core and another for .module.scss for razzle-plugin-scss.

nimaa77:
I think you have to change this config in `razzle-plugin-scss` package!

astranavt:
In modern web development we have to use preprocessors one way or another. CssModules proven well Why not include it in the engine? This will save time on the extension of the config and reduce the likelihood of errors

fivethreeo:
Yes, but you added scss as a dependency to razzle core that may not compile on some archs. Also if this is a plugin it can have more frequent updates independent of razzle core. I would recommend either separate the scss and css parts to scss-plugin and core or just add it to the scss-plugin so you can opt in on css modules, imho css modules is a anti pattern that makes large bundles and only needed if you mix different css ui libraries. But opt in is better than not being able to opt out. Plugins seems to be the preferred way of adding features to razzle.

astranavt:
I agree with you that the use of the plugin will release frequent updates and this is a very strong argument. 😀
css modules will be used only for those files that contain in the name *.module.css or *.module.scss
Create-react-app works on the same principle, and contains it all in the core
Thus, the developer has a choice of what to use, pure css or cssModules
At the moment there is an initiative to [add cssModules in chrome](https://www.chromestatus.com/feature/5948572598009856). Let's see how this direction will develop

nimaa77:
closed in favor of #1114

PR #1050, opened 2019-07-09
[Add - Example]: with "nanoexpress"

dalisoft:
This PR adds example of `razzle + nanoexpress` combo which in result will be faster response + quicker SSR

### DIFF
--- a/examples/with-nanoexpress/package.json
+++ b/examples/with-nanoexpress/package.json
@@ -1,0 +1,19 @@
+{
+  "name": "razzle-examples-with-nanoexpress",
+  "version": "3.0.0",
+  "license": "MIT",
+  "scripts": {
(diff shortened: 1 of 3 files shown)


dalisoft:
@jaredpalmer Can you look PR? I added example of nanoexpress (my backend library) + razzle combo

adaam2 (commented):
The example looks fine, although I don't really see what value it adds as it is very similar to a standard express setup.

dalisoft:
Hi and thanks for reply. 

This is similar to Express, but not same. There some things different and always having example is better than user searches. 

You may not merge, i just want add my library combo example, if possible.

fivethreeo:
nanoexpress currently inactive and deprecated

PR #1051, opened 2019-07-12
Fix type error on express.listen()

TintypeMolly:
I created an razzle app with `--example with-typescript`, built the app right after creation, and failed to build.

`const port = process.env.PORT || 3000;` makes type of `port` to be `string | 3000`.

`express.listen()` has several signatures, two of which are `listen(port: number, callback?: (...args: any[]) => void): http.Server;` and `listen(handle: any, listeningListener?: () => void): http.Server;`.

I think `port` should be of `number` type, and `express.listen()`should be inferred with the former signature.

But, for now, `port` is of an undesirable type and it gives an error which says that its arguments does not fit with the latter signature.

### DIFF
--- a/examples/with-typescript/src/index.ts
+++ b/examples/with-typescript/src/index.ts
@@ -16,7 +16,7 @@ if (module.hot) {
   console.info('✅  Server-side HMR Enabled!');
 }
 
-const port = process.env.PORT || 3000;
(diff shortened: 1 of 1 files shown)


kkalavantavanich:
Agreed that this is a good fix for the `with-typescript` example.

PR #1058, opened 2019-07-17
Update resolved dependencies to work with Node 12

kkalavantavanich:
Should fix failing CircleCI build for Node 12.


kkalavantavanich:
Passes local tests against Node 8, Node 10, and Node 12.

PR #1080, opened 2019-08-14
Added handling for user specified --watchAll

### LINKED ISSUE #1016: razzle test: --watch is not supported without git/hg
**Current behavior:** 
When using razzle in a project without a git or hg repository (we are using svn), Jest watch fails with the error code message: 
```
> razzle test --env=jsdom
Determining test suites to run...
--watch is not supported without git/hg, please use --watchAll
```

**Suggested solution:**
I tracked the issue to line 33 of /scripts/test,js. 

```
if (!process.env.CI && argv.indexOf('--coverage') < 0) {
  argv.push('--watch');
}
```

Instead of trying to determine if the files should be watch, we might think about giving that control back to the user by changing the npm script to `"test": "razzle test --env=jsdom --watch"`

xavierzwirtz:


nimaa77 (approved):
LGTM

nimaa77:
Thanks 😎

PR #1088, opened 2019-08-17
moduleResolution intent is "Node" not "node" per doc

do917:
In https://www.typescriptlang.org/docs/handbook/compiler-options.html is says default for `--moduleResolution` is `module === "AMD" or "System" or "ES6" ? "Classic" : "Node"`

In current tsconfig.json, it is lower cased "node".  Don't think this causes any unintended side effect right now but would be nice if we could be deliberate on setting it to "Node" in case.  Also, in typescript repo `protocol.d.ts` says 
`const enum ModuleResolutionKind {
      Classic = "Classic",
      Node = "Node"
  }`


jaredpalmer:
TIL.

PR #1089, opened 2019-08-17
update typescript example readme jest setting

do917:
readme has older jest transform setting.  example in `package.json` follows new requirement by `ts-jest`.  updated readme to reflect that.


fivethreeo:
Typescript example modified, closing

PR #1106, opened 2019-09-23
fix flash of unstyled content (FOUC)

nimaa77:
I was reading comments of `createConfig.js` and then I noticed this:

```javascript
// Extract our CSS into a files.
new MiniCssExtractPlugin({
  filename: 'static/css/bundle.[contenthash:8].css',
  chunkFilename: 'static/css/[name].[contenthash:8].chunk.css',
  // allChunks: true because we want all css to be included in the main
  // css bundle when doing code splitting to avoid FOUC:
  // https://github.com/facebook/create-react-app/issues/2415
  allChunks: true,
}),
```

`allChunks: true` means don't split CSS files, just put all in one file. but it's not working because there is no such a config in `MiniCssExtractPlugin`! based on MiniCssExtractPlugin [documentation](https://webpack.js.org/plugins/mini-css-extract-plugin/#extracting-all-css-in-a-single-file) we have to change `cacheGroups` in webpack config so I've changed it.

## Before merging this PR, think about config

```javascript
splitChunks: {
  cacheGroups: {
    styles: {
      name: 'bundle',
      test: /\.css$/,
      chunks: 'all',
      enforce: true,
    },
  },
},
```

this config only works with `.css` files, but what if we want to merge other files (sass or less) into `bundle.css` file?
to add support for them we can do
```javascript
test: /\.(css|scss|sass|less|styl)$/,
```
or we can ignore this, and add file extensions on plugins (like `razzle-plugin-scss`).
I think we have to do both, maybe someone updates razzle but don't update that plugin or Vice Versa

### DIFF
--- a/packages/razzle/config/createConfig.js
+++ b/packages/razzle/config/createConfig.js
@@ -457,8 +457,7 @@ module.exports = (
         new webpack.DefinePlugin(dotenv.stringified),
         // Extract our CSS into a files.
         new MiniCssExtractPlugin({
-          filename: 'static/css/bundle.[contenthash:8].css',
-          chunkFilename: 'static/css/[name].[contenthash:8].chunk.css',
+          filename: 'static/css/[name].[contenthash:8].css',
         }),
         new webpack.HashedModuleIdsPlugin(),
         new webpack.optimize.AggressiveMergingPlugin(),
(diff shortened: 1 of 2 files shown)


nimaa77:
with webpack 4, it's a little buggy

PR #1108, opened 2019-09-25
Pass environment information to modifyBabelOptions

### LINKED ISSUE #892: how to access BUILD_TARGET in babel.config.js
in order to support `babel-plugin-import` plugin ,I need to give different option depends on BUILD_TARGET
```
## for server
['import', { libraryName: 'antd'}]
## for client
['import',{libraryName: 'antd', style: true}]
```
but it seems that process.env.BUILD_TARGET is undefined in babel.config.js, how can I use BUILD_TARGET in babel.config.js

klarstrup:
Currently deriving a per-environment babel configuration is pretty tedious, involves having to do some annoying finding and mutating on the babel loader within `modify`.

This change passes the same environment information(`target` and `dev`) to `modifyBabelOptions` as is passed to `modify`, and now modifying the config for every environment is much more convenient.

### DIFF
--- a/packages/razzle/config/createConfig.js
+++ b/packages/razzle/config/createConfig.js
@@ -44,6 +44,13 @@ module.exports = (
   },
   webpackObject
 ) => {
+  // Define some useful shorthands.
+  const IS_NODE = target === 'node';
+  const IS_WEB = target === 'web';
+  const IS_PROD = env === 'prod';
+  const IS_DEV = env === 'dev';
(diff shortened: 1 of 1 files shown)


nimaa77:
we can also use env variables ( `process.env. BUILD_TARGET`) for this.
also, this is an undocumented feature :)
but it's good 👍

nimaa77 (approved):

PR #1131, opened 2019-10-22
Updated with-typescript example readme

rvanlaarhoven:


rvanlaarhoven:
Actually, after looking into the source of the `with-typescript` example, I noticed the readme is outdated. Especially the part about `razzle.config.js` does not match the contents of the file. I guess it's better to point users to https://github.com/jaredpalmer/razzle/tree/master/packages/razzle-plugin-typescript to customize the config.

fivethreeo:
Example is updated, closing

PR #1143, opened 2019-10-28
add support for proxying the devserver

fivethreeo:
Would be great to havet if you want to show your work to clients.

Currently I use replace-loader to replace the port by regular expressions.


fivethreeo:
Proof of concept https://github.com/fivethreeo/razzle-plugin-proxy

fivethreeo:
Better to upgrade webpack-dev-server

PR #1167, opened 2019-11-22
Fix vue-router error + SSR 404

wickedst:
Fixed breaking error due to vue-router update explained here: https://github.com/vuejs/vue-router/issues/2881

Also only the 404 page was being loaded on server in `vueMarkup` – replacing `req.url` with `req.originalUrl` fixed that (not sure if this is optimal but it works for now)


nimaa77:
GG thanks

PR #1178, opened 2019-12-25
Chunks plugin

nimaa77:
new chunks plugin that creates `chunks.json` file in the build directory.
contents of the file will be like: 
```js
{
  "client": {
    "css": [
      "/static/css/bundle.36d04d42.css"
    ],
    "js": [
      "/static/js/bundle.d3433edb.js",
      "/static/js/bundle.d3433edb.js.map"
    ]
  },
  "home": {
    "css": [
      "/static/css/home.2f97ec2d.chunk.css"
    ],
    "js": [
      "/static/js/home.82beecb0.chunk.js",
      "/static/js/home.82beecb0.chunk.js.map"
    ]
  },
  "about": {
    "css": [],
    "js": [
      "/static/js/about.1e639ac9.chunk.js",
      "/static/js/about.1e639ac9.chunk.js.map"
    ]
  }
}
```

this file is useful for code splitting and SSR, we can determine which chunk is for which request before we start to call `renderToString`. similar to [Collecting chunks in Loadable Components](https://loadable-components.com/docs/server-side-rendering/#collecting-chunks) but the approach I took in this PR is faster since there is no need to wait for async data (`getInitialProps`) and `renderToString` to do all their job. the only thing that we need is `req.url` and `chunkName` so we can find JS and CSS chunks for the current request.

### How It's Faster?
In SSR world we usually do two things: 
1) fetch data before start rendering 
2) render to string (HTML)

this process can take about half a second.

we can be smart and use this gap (from the moment request fired to the time which the first byte received by the browser) and tell the browser to download CSS and JS files for the current request while the server is handling the request.

other solutions like [react-loadable](https://github.com/jamiebuilds/react-loadable#mapping-loaded-modules-to-bundles) and [loadable-components](https://loadable-components.com/docs/server-side-rendering/#collecting-chunks) can't do this.

[Read this comment for more inforamtion](https://github.com/jaredpalmer/after.js/pull/237#issuecomment-536197799)

### How it's working?
1) we need to add webpackChunkName to our async components, we should also store chunkName in a variable so we can find the chunk in the next steps.

```js
// routes.js

import Home from './Home';

export default [
  {
    path: '/',
    exact: true,
    component: Home,
  },
  {
    path: '/about',
    exact: true,
    component: () => import(/* webpackChunkName: "whatever" */ './About'),
    chunkName: 'whatever',
  },
  {
    path: '/contact-us',
    exact: true,
    component: () => import(/* webpackChunkName: "ContactUs" */ './Contact'),
    chunkName: 'ContactUs',
  },
];
```
_you can develop a babel plugin that do this for you too, [like this plugin](https://github.com/jaredpalmer/after.js/blob/bc254af4578b323679a7fb3cea41d3579a535f48/packages/babel-plugin-after/README.md)._

2) find out the current request is for which component. I used react-router so I will call the `matchPath` method.
```js
const matched = matchPath(req.url, routes)
```
so `matched` variable will contain an object like this:
```js
{
  path: '/contact-us',
  exact: true,
  component: () => import(/* webpackChunkName: "ContactUs" */ './Contact'),
  chunkName: 'ContactUs',
}
```
3) `getAssets()` this is simple a function that returns a property from an array 
```js
const { scripts, styles } = getAssets(matched.chunkName, chunks)

function getAssets(chunkName, chunks) {
  const scripts = chunks[chunkName].js || [];
  const styles = chunks[chunkName].css || [];
  return { scripts, styles };
}
```

it's that much EZ :)

you can find the implemented version of the above code at [https://github.com/jaredpalmer/after.js/pull/237](https://github.com/jaredpalmer/after.js/pull/237).

### TODO
1) update all examples with the new chunk assets

@jaredpalmer check if variable names and changes are good. if everything is ok just call me to update examples.


nimaa77:
I'm going to merge this, later we can update the examples.

if I update the examples right now no one can't use it since we are not going to release a new version of razzle today

andycloke:
@nimaa77 did you add an example of this? I don't totally follow ^ so an example with TypeScript would be really useful. I can't see one in the examples folder but I may well be missing it!

nimaa77:
@andycloke oh sorry didn't notice your message here :)

checkout https://github.com/jaredpalmer/after.js we are using it mostly in that library

PR #1185, opened 2020-01-24
Updating includes to be future-safe, fixing the suggestions in warnings.

najamkhn:
This change is a fix for the following warnings below:

```
Warning: Please use `require("react-router-dom").Route` instead of `require("react-router-dom/Route")`. Support for the latter will be removed in the next major release.
Warning: Please use `require("react-router-dom").Switch` instead of `require("react-router-dom/Switch")`. Support for the latter will be removed in the next major release.

```

### DIFF
--- a/examples/with-react-loadable/src/App.js
+++ b/examples/with-react-loadable/src/App.js
@@ -1,6 +1,5 @@
 import React from 'react';
-import Route from 'react-router-dom/Route';
-import Switch from 'react-router-dom/Switch';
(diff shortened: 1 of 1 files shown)


nimaa77:
@najamkhn hello, I saw the example and there are some other components that import in this way, please update those as well and then run prettier ...

nimaa77:
I think Github won't send me an email or a notification when you push on this, just mention me once everything was ok.

nimaa77:
closed in favor of #1278

PR #1202, opened 2020-03-02
add with-emotion example

tronjsdev:
Just another example that using emotionjs 11


fivethreeo:
Could you do this against the canary branch?

PR #1217, opened 2020-03-20
Fixed windows test in github action

wickedev:
I did the following to fix the windows test:

- testMatch glob pattern change for windows
- github action yarn install network timeout change to 20s
- e2e test fix on windows by path.join


wickedev:
Close to rework with the master branch target.

PR #1218, opened 2020-03-20
Fix windows tests in Github Action and Circle CI

wickedev:
I did the following to fix the test:

- testMatch glob pattern change for windows (https://github.com/facebook/jest/issues/7914#issuecomment-464352069)
- github action yarn install network timeout change to 30s
    - I don't know why, but when fetch the resolve-url-loader on windows github action using yarn, it is timed out. upgrading to the latest version will fix the problem, but there is a risk, so added the `--network-timeout 30000` flag instead upgrading to the latest version.
- e2e test fix on windows by path.join
- bs-platform dependency of examples/with-reason-react causes intermittent problems in the 'Install and bootstrap packages' phase of circle ci, so it was changed to the latest version. As it is an example, the risk is expected to be small.

### DIFF
--- a/examples/with-reason-react/package.json
+++ b/examples/with-reason-react/package.json
@@ -18,7 +18,7 @@
     "reason-react": "^0.5.3"
   },
   "devDependencies": {
-    "bs-platform": "^4.0.17",
+    "bs-platform": "^7.2.2",
     "concurrently": "^4.0.1",
     "razzle": "^2.4.1"
   }
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -25,7 +25,7 @@ jobs:
       run: yarn global add elm
 
     - name: Install and bootstrap packages
-      run: yarn install --frozen-lockfile --ignore-engines
+      run: yarn install --frozen-lockfile --ignore-engines --network-timeout 30000
 
     - name: Run tests
       run: yarn e2e --runInBand --coverage
--- a/test/jest.config.json
+++ b/test/jest.config.json
@@ -2,6 +2,6 @@
   "roots": ["<rootDir>/tests"],
   "collectCoverageFrom": ["**/*.js"],
   "testMatch": [
-    "<rootDir>/tests/**/?(*.)(spec|test).(ts|js)?(x)"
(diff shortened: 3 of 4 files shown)


wickedev:
When I send a PR from a my forked repository, the CI does not run automatically, so I will refer to the CI report in my repository.

- CircleCI: https://circleci.com/gh/wickedev/razzle
- Github Action: https://github.com/wickedev/razzle/actions/runs/59534931

nimaa77:
hey @wickedev thanks nice job

what is the problem with upgrading`resolve-url-loader` if it doesn't drop support for node 8 we can upgrade it to lasted version

examples/with-reason-react/package.json
```diff
@@ -18,7 +18,7 @@
     "reason-react": "^0.5.3"
   },
   "devDependencies": {
-    "bs-platform": "^4.0.17",
+    "bs-platform": "^7.2.2",
```
nimaa77, line 21: could you please explain why you upgraded this package?

wickedev: [In the 'Install and bootstrap packages' phase of the Circle CT 'test-node9' target, the bs-platform build fails](https://circleci.com/gh/jaredpalmer/razzle/1313). I'm not sure why, but it was fixed by upgrading the bs-platform version. Another solution is to drop non LTS versions of the tests. Problem case of the module currently being used by razzle is get-caller-file, which prevents the build from breaking through the --ignore-engines flag, but it doesn't seem to be the underlying solution. https://github.com/stefanpenner/get-caller-file/issues/18

wickedev:
> what is the problem with upgrading`resolve-url-loader` if it doesn't drop support for node 8 we can upgrade it to lasted version

After upgrading version of `resolve-url-loader` to ^3.1.1, `yarn link razzle-plugin-scss` was executed in`examples/with-scss`. start, test, build, and start: prod all command seem to work fine. However, since the test of the `razzle-plugin-scss` package was originally broken, I do not know if there is a problem with the package self-test. If you think it's better to upgrade the version than `--network-timeout 30000`, I will fix that.

nimaa77:
@wickedev can we merge this?

wickedev:
@nimaa77 Sure. No problem.

PR #1222, opened 2020-03-25
Plugins from organisations

Ekman:
We have an npmjs.org organisation at work. We create packages with names like `@myorg/name`, see: https://docs.npmjs.com/about-org-scopes-and-packages. In an eslint configuration you can write `@myorg` and it will automatically find a package named `@myorder/eslint-config`, see: https://eslint.org/docs/developer-guide/shareable-configs#npm-scoped-modules.

Since we have multiple Razzle applications, we have a Razzle plugin in order to centralize most of our build configuration to one place. The name of that package is `@myorg/razzle-plugin`. This change will allow the same naming convention for Razzle plugins as eslint plugins.

Let me know if you have any feedback.


fivethreeo:
added in finch branch

PR #1226, opened 2020-04-02
Example with navigation and digitalocean deployment instructions

strommj:
I created a simple example of utilizing a Navigation bar via React-Router.  Also included in here are instructions to deploy to a DigitalOcean droplet.


strommj:
It seems the yarn.lock file needs updating, but I'm not sure it's best to include that in this PR.  I confirmed this by freshly cloning the project and running `yarn`, and confirmed it was different from the repository version.

strommj:
Is there a particular reason why you closed this PR?

fivethreeo:
Ah, sorry did not see the README 😀

strommj:
Sorry I was inactive on this PR, I updated the yarn.lock file to match master to eliminate any changes from my local build. I also updated the project to use razzle@4.0.4

PR #1235, opened 2020-04-09
fix loadable-components example

fivethreeo:


sedlukha:
There is a good example in loadable-components repo and it works great.  https://github.com/gregberge/loadable-components/tree/master/examples/razzle

fivethreeo:
Yes, maybe I should try getting that one in instead.

PR #1240, opened 2020-04-12
promisify building/modifying configs and paths, add extra confighooks 

fivethreeo:


fivethreeo:
Needs some fixing wrt. env and code readability. Possibly taking away unnecessary hooks or combining some hooks.

nimaa77:
@fivethreeo I don't know why this is useful ... (it's lack of my knowledge)

but we can ask @jaredpalmer to have a review on this

fivethreeo:
I am not quite happy with this solution yet. But the case for this is to allow async code to modify the config and the configOptions part is to avoid complex ways of changing simple things in the config already created, changing paths can be useful too. E.g. create ssl certificates before continuing the configuration. But the api is not quite the way I want it to be yet. Also it could be useful to have some hooks into createJestConfig.

fivethreeo:
I'll ping when I have made some api changes. :)

fivethreeo:
I'll close this and do it more incrementally, too many changes at once. I'll do asyc/await first. Then paths. Current idea adds more complexity and that is not what razzle is about.

PR #1275, opened 2020-04-26
Handle typescript with Babel

### LINKED ISSUE #996: Disable type checking when using typescript
Hello!

Is there any way to use typescript with razzle, but without all the type checking?

`transpileOnly = true` by default and I tried setting `forkTsChecker.typeCheck = false`, but type checking still fires.

I found I can silence the logger, but is there any way to disable checking altogether? I'm using VSCode, which does that already.

Thank you!

### LINKED ISSUE #1242: Updating Plugins, And Removing Typescript plugin!!
I'm going to update these plugins (some of them are not implemented yet and some other very old)
so we are going to have some breaking changes

- razzle-plugin-mdx
- razzle-plugin-preact
- razzle-plugin-scss
- razzle-plugin-typescript (maybe remove this??)
- razzle-plugin-eslint

### What? Why We Remove Typescript Plugin?
We are using `babel` for transpiling our Javascript code right now. recently (about a year ago) babel team and typescript team released a new plugin called [babel-plugin-transform-typescript](https://babeljs.io/docs/en/babel-plugin-transform-typescript) so with this plugin babel can transpile Typescript code to Javascript.

in:
```ts
const x: number = 0;
```

out:
```js
const x = 0;
```

#### What about type checking

just install `tsc` (typescript compiler) and we are good to go:
```
yarn add tsc
yarn tsc
```

#### What about configuration?
create a valid `tsconfig.json` at the root of your project and that's all.

#### What about absolute imports?
WE ALREADY COVERED THAT 😎
https://github.com/jaredpalmer/razzle/pull/1112

so we can add `babel-plugin-transform-typescript` to `babel-preset-razzle` and also update the webpack config to handle ts and tsx files with babel-loader:

```js
{
  test: ts|tsx|js|jsx$,
  loader: babel-loader
}
```

with these changes, adding typescript to razzle is going to be super easy, only in two steps:
1. `yarn add tsc`
2. create `tsconfig.json` at the root of the project

piecyk:
* [x] Add `@babel/preset-typescript` to `babel-preset-razzle`
* [x] Recognise `.ts` `.tsx` extensions in razzle
* [x] Update with-typescript example 
* [x] ~Remove razzle-plugin-typescript ( extracted from #1242 )~ reverted  
* [x] Remove with-redux-observable example ( always we can add it back, when someone has time to fix it )
* [x] ~Update razzle-plugin-typescript to use only fork-ts-checker-webpack-plugin~ no need, current approach with babel with not break ts-loader setup 
* [x] Add note about depreciation of `razzle-plugin-typescript`


nimaa77:
+1 for removing "with-redux-observable" 

that example is crazy

piecyk:
> that example is crazy

Exactly, did a quick look on it, but after a sec... yeah let's not go there 😅 

Also need to re-create `yarn.lock` as it was throwing with some weird error when adding @babel/preset-typescript

nimaa77:
sure I'm going to check this in 15 minutes

examples/with-typescript/package.json
```diff
@@ -3,31 +3,29 @@
...
-    "tslint-react": "^4.0.0",
-    "typescript": "^3.5.3"
+    "@types/express": "^4.17.6",
+    "@types/jest": "^25.2.1",
+    "@types/node": "^13.13.2",
+    "@types/react": "^16.9.34",
+    "@types/react-dom": "^16.9.6",
+    "@types/react-router-dom": "^5.1.4",
+    "@types/webpack-env": "^1.15.2",
+    "razzle": "^3.1.0",
```
nimaa77:
```suggestion
    "razzle": "latest",
```

nimaa77:
but wait, is this a breaking change?

piecyk:
Ooo true, yeah let's revert removing `razzle-plugin-typescript` but then we should update the plugin to don't use ts-loader, as it's not needed, only the fork-ts as CRA is doing. But would prefer to have the example without the plugin as it's now. 

Should i update the plugin or are you doing it already ?

nimaa77:
I'm working on it

nimaa77:
I'm still thinking about the breaking change
what if someone who already uses razzle updates to this version and it's app stop working because the transpiler is changed !!

piecyk:
Basic need to update, `razzle` and `razzle-plugin-typescript` together, imho it's not a breaking change.

nimaa77:
@fivethreeo what's your idea

nimaa77:
I think it's breaking change 
https://babeljs.io/docs/en/babel-plugin-transform-typescript#caveats

nimaa77:
example is good

nimaa77:
we should keep the new example (changing the example is not a breaking change)

nimaa77:
ok we have two solutions
1) add this feature under a flag (if the flag was set we add .ts and .tsx extensions to babel-loader) and everything is going to work well

2) in `razzle-plugin-typescript` we remove `.ts` and `.tsx` extentions of babel-loader

```js
// Safely locate Babel-Loader in Razzle's webpack internals
const babelLoader = config.module.rules.find(babelLoaderFinder);

babelLoader.test = test: /\.jsx?$/, 
```

nimaa77:
also found this in `razzle-plugin-typescript`, so if `useBabel` is disabled (which is disabled by default) it's removing babelLoader completely

```js
  if (options.useBabel) {
    // If using babel, also add babel-loader to ts files,
    // so we can use babel plugins on tsx files too
    tsLoader.use = [...babelLoader.use, ...tsLoader.use];
  } else {
    // If not using babel, remove it
    config.module.rules = config.module.rules.filter(
      rule => !babelLoaderFinder(rule)
    );
  }
```

nimaa77:
this PR is not going to breaking anything :)

webpack composes `loader.use`, this means it's going to execute them from right to left.

so tsLoader will transpile Typscript to Javascript and then pass it to babel ...
```js
tsLoader.use = [...babelLoader.use, ...tsLoader.use];
```

### just in case
in typescript-plugin, I'm going to add `exclude` for `.ts` and `.ts` to babel-loader ...

nimaa77:
ok, I think we don't need to add `ts-jest` because babel is transpiling Typescript to Javascript ...

nimaa77:
I also changed the `start` script
if you think it's better to have `tsc` and `razzle start` in one command just revert my commit 

and also double-check and see if everything is working, if you had time just check if typescript-plugin is working and it has the `exclude` option in right place ...

(I'm really tired sorry) 

once everything was ok, ping me to merge this awesome pull request

piecyk:
@nimaa77 don't by sorry, great work 👏 

> if you think it's better to have tsc and razzle start in one command just revert my commit

Yep i think that running type-checking with build in watch is essential for dev experience and must have. Also as we removed `ts-jest` we can move the whole configuration to `createJestConfig.js` what i did 😄Plus added `@testing-library` similar as it's in CRA as it's now an industry standard.


> this PR is not going to breaking anything :) 

🎉 awesome, imho we should add depreciation that it will be removed in next major version. 
Plus add a note that `with-typescript` example is official way using off using typescript with razzle.

Something is wrong with `razzle-plugin-typescript`, when trying to run test it throws `When you use tslint option, make sure to install tslint` 🤔 Would not spent to much time on it as `tslint` is deprecated and current approach will not break.

fivethreeo:
Seems like you guys got a good grip on ts configuration. From what I get this solution seems good. Will ts be typechecked in watch mode?

nimaa77:
yes

nimaa77 (approved):

fivethreeo:
Working on a github action that can deprecate packages via a console script that can be run with a github api key. So we can do a npm deprecate via github. Is it just deprecate in razzle-plugin-typescript readme that is missing here. I can add that and merge this?

fivethreeo:
Added deprecation note. Merged, great work :)

PR #1280, opened 2020-04-28
Use `setupFilesAfterEnv` in Jest Config

rclmenezes:
Jest is printing out a deprecation warning:

> Option "setupTestFrameworkScriptFile" was replaced by configuration "setupFilesAfterEnv", which supports multiple paths.

See documentation on `setupFilesAfterEnv` here: https://jestjs.io/docs/en/configuration#setupfilesafterenv-array

### DIFF
--- a/packages/razzle/config/createJestConfig.js
+++ b/packages/razzle/config/createJestConfig.js
@@ -25,7 +25,7 @@ module.exports = (resolve, rootDir) => {
   // in Jest configs. We need help from somebody with Windows to determine this.
   const config = {
     collectCoverageFrom: ['src/**/*.{js,jsx,mjs}'],
-    setupTestFrameworkScriptFile: setupTestsFile,
+    setupFilesAfterEnv: [setupTestsFile],
     testMatch: [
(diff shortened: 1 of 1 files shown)


nimaa77:
@fivethreeo I think this is a Breaking Change

fivethreeo:
But should we upgrade jest, and put this in next? Or is that a 3.3 release if it is a breaking change.

hoop71:
> But should we upgrade jest, and put this in next? Or is that a 3.3 release if it is a breaking change.

Jest just went to 26. A few breaking changes in there as well. https://github.com/facebook/jest/blob/master/CHANGELOG.md

octavianparalescu:
Hello guys,
I am receiving the same warning after upgrading the packages in our project, I think it would be great to merge this PR so we can continue using the default test script to run our tests.
Thanks

fivethreeo:
You can try out the canary release. 

```bash
yarn add razzle@canary
```

fivethreeo:
fixed in dev

PR #1283, opened 2020-05-04
Create razzle app improved

fivethreeo:
# use default template
packages/create-razzle-app/bin/create-razzle-app cra

# use official example
packages/create-razzle-app/bin/create-razzle-app --example=basic cra

# use example from npm
packages/create-razzle-app/bin/create-razzle-app --example=razzle-example-basic cra
packages/create-razzle-app/bin/create-razzle-app --example=razzle-example-basic@latest cra

# use example in subdir from npm
packages/create-razzle-app/bin/create-razzle-app --example=razzle-example-basic:subexample cra
packages/create-razzle-app/bin/create-razzle-app --example=razzle-example-basic@latest:subexample cra

# use example from github
packages/create-razzle-app/bin/create-razzle-app --example=https://github.com/fivethreeo/razzle-example-basic cra
packages/create-razzle-app/bin/create-razzle-app --example=https://github.com/fivethreeo/razzle-example-basic@master cra

# use example in subdir from github
packages/create-razzle-app/bin/create-razzle-app --example=https://github.com/fivethreeo/razzle-example-basic:subexample cra
packages/create-razzle-app/bin/create-razzle-app --example=https://github.com/fivethreeo/razzle-example-basic@master:subexample cra

# use example from git
packages/create-razzle-app/bin/create-razzle-app --example=git+https://github.com/fivethreeo/razzle-example-basic cra
packages/create-razzle-app/bin/create-razzle-app --example=git+https://github.com/fivethreeo/razzle-example-basic@master cra

# use example in subdir from git
packages/create-razzle-app/bin/create-razzle-app --example=git+https://github.com/fivethreeo/razzle-example-basic:subexample cra
packages/create-razzle-app/bin/create-razzle-app --example=git+https://github.com/fivethreeo/razzle-example-basic@master:subexample cra

# use example from local path
packages/create-razzle-app/bin/create-razzle-app --example=file:examples/basic cra

# real world example 
packages/create-razzle-app/bin/create-razzle-app --example=https://github.com/gregberge/loadable-components:examples/razzle cra

### DIFF
--- a/test/tests/create-razzle-app.test.js
+++ b/test/tests/create-razzle-app.test.js
@@ -1,0 +1,212 @@
+/**
+ * @jest-environment node
+ */
+'use strict';
+
+const shell = require('shelljs');
+const util = require('../fixtures/util');
+const path = require("path");
+
+shell.config.silent = false;
+
+const stageName = 'stage-cra';
+const craPath = path.join('../node_modules/.bin/create-razzle-app');
+const testPackage = 'razzle-example-basic';
+const testRepo = 'https://github.com/fivethreeo/razzle-example-basic';
+
+describe('create-razzle-app', () => {
+  beforeAll(() => {
+    util.teardownStage(stageName);
+  });
+
+  it('should create app from default template', () => {
+    util.setupStage(stageName);
+    const output = shell.exec(
+      `${craPath} cra --no-install`
+    );
+    expect(shell.test('-d', 'cra/node_modules')).toBeFalsy();
+
+    expect(shell.test('-f', 'cra/yarn.lock')).toBeFalsy();
+
+    expect(shell.test('-f', 'cra/package.json')).toBeTruthy();
+
+    expect(shell.test('-d', 'cra/src')).toBeTruthy();
+    expect(shell.ls('cra/src/index.js').code).toBe(0);
+
+    expect(output.code).toBe(0);
+  });
+
+  it('should create app from default template and install packages', () => {
+    util.setupStage(stageName);
+    const output = shell.exec(
+      `${craPath} cra`
+    );
+    expect(shell.test('-d', 'cra/node_modules')).toBeTruthy();
+
+    expect(shell.test('-f', 'cra/yarn.lock')).toBeTruthy();
+
+    expect(shell.test('-f', 'cra/package.json')).toBeTruthy();
+
+    expect(shell.test('-d', 'cra/src')).toBeTruthy();
+    expect(shell.ls('cra/src/index.js').code).toBe(0);
+
+    expect(output.code).toBe(0);
+  });
+
+  it('should create app from official example', () => {
+    util.setupStage(stageName);
+    const output = shell.exec(
+      `${craPath} cra --example=basic --no-install`
+    );
+    expect(shell.test('-f', 'cra/package.json')).toBeTruthy();
+
+    expect(shell.test('-d', 'cra/src')).toBeTruthy();
+    expect(shell.ls('cra/src/index.js').code).toBe(0);
+
+    expect(output.code).toBe(0);
+  });
+
+  it('should create app from npm example', () => {
+    util.setupStage(stageName);
+    const output = shell.exec(
+      `${craPath} cra --example=${testPackage} --no-install`
+    );
+    expect(shell.test('-f', 'cra/package.json')).toBeTruthy();
+
+    expect(shell.test('-d', 'cra/src')).toBeTruthy();
+    expect(shell.ls('cra/src/index.js').code).toBe(0);
+
+    expect(output.code).toBe(0);
+  });
+
+  it('should create app from npm example at tag', () => {
+    util.setupStage(stageName);
+    const output = shell.exec(
+      `${craPath} cra --example=${testPackage}@latest --no-install`
+    );
+    expect(shell.test('-f', 'cra/package.json')).toBeTruthy();
+
+    expect(shell.test('-d', 'cra/src')).toBeTruthy();
+    expect(shell.ls('cra/src/index.js').code).toBe(0);
+
+    expect(output.code).toBe(0);
+  });
+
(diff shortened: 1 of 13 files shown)


sergeysova:
Conflicts

fivethreeo:
i was thinking just not have "official" example and just use the github example loading with subdir for the official ones too, opinions?

nimaa77:
@fivethreeo can we add old `--examples` option!
that option had a nice developer experience you only need to know what you want and you just type the name

```bash
npx create-razzle-app --example=with-redux my-redux-ssr-app
```

but using full URL is kind of boring (at least for official examples)

fivethreeo:
Yes but it requires a extra step in releases, see  scripts/create-examples-list.js

fivethreeo:
But if we remove npm examples, that won't be needed.

nimaa77:
we don't need npm examples they can directly install examples from npm, with `npm install fooBarBaz`

fivethreeo:
Ok, then I know what to do :)

sergeysova:
@nimaa77 I don’t need to install example as npm package. I want to create my own example. I want to use it as template to bootstrap new application.

```shell
yarn create razzle-app --template effector my-app
```

fivethreeo:
@sergeysova good argument, will consider it

fivethreeo:
Looking a bit closer at this I see create-react-app uses examples from npm so I think we should too.

fivethreeo:
create-react-app also supports file:path to load a local example

fivethreeo:
So I think we keep the old way of loading official examples, then add github, git, npm and file: support.

packages/create-razzle-app/lib/examples.js
```diff
@@ -0,0 +1 @@
+module.exports = ["basic","basic-spa","with-afterjs","with-custom-babel-config","with-custom-devserver-options","with-custom-environment-variables","with-custom-webpack-config","with-elm","with-firebase-functions","with-heroku","with-hyperapp","with-inferno","with-jest-snapshots","with-jsxstyle","with-koa","with-less","with-loadable-components","with-material-ui","with-mdx","with-now","with-now-v2","with-polka","with-preact","with-rax","with-react-loadable","with-react-native-web","with-react-router-3","with-reason-react","with-redux","with-redux-observable","with-scss","with-styled-components","with-typescript","with-vendor-bundle","with-vue","with-vue-router"];
```
nimaa77: it's not good to hard code example names
why this can't be dynamic?

test/tests/create-razzle-app.test.js
```diff
@@ -0,0 +1,191 @@
...
+const path = require("path");
+
+shell.config.silent = false;
+
+const stageName = 'stage-cra';
+const craPath = path.join('../node_modules/.bin/create-razzle-app');
+const testPackage = 'razzle-example-basic';
+const testRepo = 'https://github.com/fivethreeo/razzle-example-basic';
+
+describe('create-razzle-app', () => {
```
nimaa77, line 17: can we add a test and check if `node_modules` directory is available or not?
so basically when we run the command with `--no-install` we expect that node_modules folder is empty (or it's not get created from first place) we can also check for `yarn.lock` or `package-lock.json` file ...

nimaa77 (approved):

test/tests/create-razzle-app.test.js
fivethreeo: Maybe just do that to one test? Npm takes so much time to run.

packages/create-razzle-app/lib/examples.js
nimaa77: we can use Axios with GitHub API to check if one example is available or not

```js
const repoOwner = "jaredpalmer";
const repoName = "razzle";
// please don't use start the path with "/" (or maybe we handle this in the URL...)
const examplesDirectoryPath = "examples";

const isFolder = ({ type }) => type === "dir";
// a little functional programming 
const prop = (key) => (obj) => obj[key];

const { data } = await axios.get(`https://api.github.com/repos/${repoOwner}/${repoName}/contents/${examplesDirectoryPath}`)
const validExampleNames = data.filter(isFolder).map(prop("name"))
```

and this is how `validExmapleNames` will look like at the end
```js
["basic-spa", "basic", "with-afterjs", "with-custom-babel-config", "with-custom-devserver-options", "with-custom-environment-variables", "with-custom-webpack-config", "with-elm", "with-firebase-functions", "with-heroku", "with-hyperapp", "with-inferno", "with-jest-snapshots", "with-jsxstyle", "with-koa", "with-loadable-components", "with-material-ui", "with-mdx", "with-now-v2", "with-now", "with-polka", "with-preact", "with-rax", "with-react-loadable", "with-react-native-web", "with-react-router", "with-reason-react", "with-redux", "with-scss", "with-styled-components", "with-typescript", "with-vendor-bundle", "with-vue-router", "with-vue"]
 ```

more info:
https://developer.github.com/v3/repos/contents/
https://api.github.com/repos/jaredpalmer/razzle/contents/examples

test/tests/create-razzle-app.test.js
nimaa77: yes yes yes

packages/create-razzle-app/lib/examples.js
fivethreeo: Because how do we distinguish official examples from npm examples. npm: prefix? But then we differ from create-react-app. Or move examples into create-razzle-app, but then that becomes large to load with npx. So prefix or hardcode.

nimaa77: I think it's better to add a prefix for npm
but we can call the GitHub API at first and then we look for the example if we didn't find it then we try NPM ...

fivethreeo: Good idea :)

fivethreeo:
Merged

PR #1293, opened 2020-05-11
fixed the chunks 404 error

ganeshrvel:
The script `assets[key].js` can also be an array which was not handled properly

### DIFF
--- a/examples/with-vendor-bundle/src/server.js
+++ b/examples/with-vendor-bundle/src/server.js
@@ -1,32 +1,48 @@
-import App from "./App";
-import React from "react";
-import express from "express";
-import { renderToString } from "react-dom/server";
+import React from 'react';
+import express from 'express';
+import { renderToString } from 'react-dom/server';
+import { StaticRouter } from 'react-router-dom';
+import App from './features/App';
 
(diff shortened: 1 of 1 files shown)


fivethreeo:
This pull changes unrelated things and only fixes one example.

And change the require back  to

```js
require(process.env.RAZZLE_ASSETS_MANIFEST);
```

But if you clean it up and apply it to all examples in the next branch, this stands a chance to be pulled :)

nimaa77:
I think we can use the `chunks.json` file for this.

I just tried to refactor this example but I didn't get what we are trying to achieve

fivethreeo:
What he says is not each asset is just a string.

nimaa77:
we can use `Array.prototype.flat` to flatten the array at first 
> maybe use `Set` to remove duplicated items (if any)

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat

ganeshrvel:
@fivethreeo new changes are up.

fivethreeo:
We need to look into this some more and consider updating all the examples.

fivethreeo:
pulled into dev

PR #1305, opened 2020-05-19
Fix example code documentation

slobo:
I don't think there is anything Immutable in taking a reference to an object and then modifying the reference like is done in the example of vendor chunking config file. 

We would need a deep copy for true immutability, so best to remove the misleading bit of code, eh? Or is immutability actually required here?

```
> const a = { hello: "world"}
undefined
> const b = a
undefined
> b.hello = "earth"
'earth'
> a.hello
'earth'
```

### DIFF
--- a/examples/with-vendor-bundle/razzle.config.js
+++ b/examples/with-vendor-bundle/razzle.config.js
@@ -1,9 +1,7 @@
 'use strict';
 
 module.exports = {
-  modify(defaultConfig, { target, dev }, webpack) {
-    const config = defaultConfig; // stay immutable here
-
+  modify(config, { target, dev }, webpack) {
     // Change the name of the server output file in production
     if (target === 'web') {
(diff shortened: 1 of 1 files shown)


fivethreeo (commented):
It should be const config = Object.assign({}, defaultConfig); like the other examples. How important it is I dont know.

slobo:
That still is not a deep copy, for which you would need some more involved code or a technique like JSON serialize + deserialize IIRC

```
> const a = { hello: { deep: "world"} }
undefined
> const b = Object.assign({}, a);
undefined
> b.hello.deep = 'earth'
'earth'
> a
{ hello: { deep: 'earth' } }
>
```

fivethreeo:
I think you are right 😀should update all examples and docs in next branch.

fivethreeo:
Fixed in dev

PR #1308, opened 2020-05-20
Update jest deps to 26.0.1

### LINKED ISSUE #1291: Upgrade Jest to version 26
### Current Behavior

Jest 25 utilizes jsdom 15. This has introduced a few errors in my tests that I don't have using Jest 26 outside of Razzle. I'm nearly certain this is an issue related to having an old version of Jest: https://github.com/testing-library/dom-testing-library/issues/477

### Desired Behavior

Use Jest 26 throughout the project.

### Suggested Solution

Upgrade Jest.

### Who does this impact? Who is this for?

All users I suppose.

### Additional context

Read this issue: https://github.com/testing-library/dom-testing-library/issues/477

fivethreeo:


cereallarceny (changes requested):
I believe you missed one reference to Jest needing an upgrade here: https://github.com/jaredpalmer/razzle/blob/604b5105518f630b8627fd9f463860e1078c0d9e/package.json#L15

fivethreeo:
Yes, then it blows up :) :(

cereallarceny:
Oh boy. Well so long as that's just the internal jest version, I guess that's fine. When I run `razzle test`, which Jest dependency is it running against? The main `package.json` or one of the ones in the `packages` folder?

fivethreeo:
Best would be to fix the tests.

fivethreeo:
And then this also seems like a breaking change. :(

fivethreeo:
It might have to do with some parallel testing. So need to figure out what changed in jest.

fivethreeo:
Seems like this has to be pushed to a 4.0 or possibly a 3.4 release. But solving it now won’t hurt. If you can figure out why the tests fail we can reach 4.0 faster 😀

cereallarceny:
I mean, I don't see why you wouldn't just merge the PR as is and then open a new issue for fixing the broken tests. I'll take the progress, even if it's not everything.

fivethreeo:
If it breaks our tests, it might also break others tests. So we should warn in the next release and pull in the changes after that. And having broken tests delays releases.

cereallarceny:
Do you have any planned roadmap? @jaredpalmer had this but since you’ve
taken over the project, it would be nice to see some roadmap of your
intentions.

PR #1309, opened 2020-05-20
Use contenthash for better longterm caching of JS assets

sumanbh:
## Summary
We should use contenthash instead of chunkhash for better longterm caching of JS assets. chunkhash relies on the whole chunk as opposed to contenthash which relies only on the content of the JS file itself

https://webpack.js.org/guides/caching/#output-filenames


fivethreeo:
fixed in dev

PR #1326, opened 2020-06-08
chore: upgrade assets-webpack-plugin beyond 3.10

Lapple:
To upgrade mkdirp to 1.x and remove installation warnings:

> Legacy versions of mkdirp are no longer supported. Please update to mkdirp 1.x.
> (Note that the API surface has changed to use Promises in 1.x.)

Related PR to the plugin source: https://github.com/ztoben/assets-webpack-plugin/pull/230

Does not require an immediate version release, I believe, just something good for future 😄

### DIFF
--- a/packages/razzle/package.json
+++ b/packages/razzle/package.json
@@ -17,7 +17,7 @@
   ],
   "dependencies": {
     "@babel/core": "^7.8.7",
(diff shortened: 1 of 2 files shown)


fivethreeo:
fixed in dev

PR #1345, opened 2020-07-06
Next prerender

fivethreeo:
node -e 'require("./test/fixtures/util").setupStageWithExample("example", "with-react-router", symlink=false, yarnlink=true, install=true, test=false);'
cd example
yarn build
yarn export
yarn start:static
cd ..
node -e 'require("./test/fixtures/util").teardownStage("example");'


nimaa77:
ok, I'm going to try this and see how it goes

PR #1357, opened 2020-07-14
fix(razzle-dev-utils): Missing dependency in dev utils

martinlistiak:
This missing dependency started causing problems while starting a project.

### DIFF
--- a/packages/razzle-dev-utils/package.json
+++ b/packages/razzle-dev-utils/package.json
@@ -20,6 +20,7 @@
     "chalk": "3.0.0",
     "jest-message-util": "^24.9.0",
     "react-dev-utils": "^10.2.0",
-    "strip-ansi": "6.0.0"
+    "strip-ansi": "6.0.0",
+    "webpack-dev-server": "~3.10.3"
   }
 }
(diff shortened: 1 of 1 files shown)


fivethreeo:
If anything, it should be a peerDependency with the same version as razzle core.

fivethreeo:
If you run this in the razzle-dev-utils package dir you get what should be added to package.json :)

```bash
sudo npm install dependency-check -g
sudo apt-get install moreutils

dependency-check ./package.json --verbose ./*.js --ignore 2>&1 | awk '
/Dependencies not listed/ {
  for (i=7;i<=NF;i++) {
    gsub(",", "", $i);
    print $i;
  }
}
' | xargs -n 1 yarn why 2>&1 | sponge |  awk '
BEGIN {
  FS="\"";
  out="";
} 
/Found/ {
  gsub("@", "\": \"", $2);
  out=out"    \""$2"\",\n";
}
END {
  gsub(/,\n$/, "", out);
  print "  \"peerDependencies\": {";
  print out;
  print "  }";
} 
'
```

fivethreeo:
Do you use npm or yarn in your project?

fivethreeo:
fixed in 3.1.7

PR #1362, opened 2020-07-18
doc: remove -- from yarn command for arguments

alireza-mh:
there is no need for `--` in `yarn` command to pass arguments.


fivethreeo:
fixed in dev

PR #1389, opened 2020-08-20
Added async and await to test command

Orfidius:
While trying to upgrade Razzle to a version that did not have the deprecation warning, I noticed that razzle's createJestConfig function had been changed to a promise but was not being consumed using async and await. This resulted in custom configs not being pulled in properly from the package.json. I've added async and await, and tested the test command to ensure it works properly. 

This is my first pull request on an open source project, not sure if I'm following proper procedures, but thanks for the hard work!

### DIFF
--- a/packages/razzle/scripts/test.js
+++ b/packages/razzle/scripts/test.js
@@ -45,11 +45,11 @@ const fs = require('fs-extra');
 const defaultPaths = require('../config/paths');
 
 loadRazzleConfig(webpack, defaultPaths).then(
-  ({ razzle, webpackObject, plugins, paths }) => {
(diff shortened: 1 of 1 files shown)


fivethreeo:
Thanks for your interest in async razzle. This is fixed in dev and will be in the next canary :)

fivethreeo:
https://github.com/jaredpalmer/razzle/tree/next-awesome-plugins

fivethreeo:
Your way of adding this was on point by the way. Feel free to keep sending pulls for needed changes 😃

Orfidius:
:) Thanks! I hope to submit something useful next time!

PR #1418, opened 2020-09-27
Add prefresh to example

JoviDeCroock:
Prefresh is the equivalent of react-fast-refresh but for  Preact


fivethreeo:
Can you do this against the finch branch?

fivethreeo:
Will only add fixes or security fixes to master, closing.

PR #1421, opened 2020-09-28
Add prefresh example to finch

JoviDeCroock:
In finch razzle seems to have a bug where it hard-requires React so the Preact example is broken, due to this bug I temp added React as a dependency. The odd one out is that the HMR-runtime used inside of razzle seems broken since it can never find the generated hot-update file.

### DIFF
--- a/examples/with-preact/.gitignore
+++ b/examples/with-preact/.gitignore
@@ -11,3 +11,4 @@ public/static
 .env.development.local
 .env.test.local
 .env.production.local
+cache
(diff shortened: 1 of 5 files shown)


fivethreeo:
Do you know where React is hard required? Does HMR work in this example?

JoviDeCroock:
HMR seems broken, HMR works on the original PR not on this one, will investigate later but not familiar with razzle-core yet.

JoviDeCroock:
Seems to work correctly now, can't figure out what's hard-requiring React though.

fivethreeo:
If you squash this into one commit and make the commit message "example: add preact prefresh example" i will merge it :)

PR #1467, opened 2020-11-10
Support for not released plugins in razzle.config.js

sneridagh:
Sometimes, you want to add a plugin that is not released to npm, but has the form of a local module.
You are able to do that as a function, but not using the new object shaped plugins.


fivethreeo:
merged

Issue #1466, opened 2020-11-09
with-typescript example doesn't resolve non-relative imports in test runner


parkerault:
### Current Behavior

If using non-relative imports by defining `baseUrl` in `tsconfig.json`, the start and build scripts will work, but the test runner fails when you try to import a file that has non-relative imports.

### Expected behavior

The test runner should resolve non-relative imports the same as the start and build scripts.

### Reproducible example

https://github.com/parkerault/razzle-test-imports-failure

### Additional context

For context, see @fivethreeo's comments here: https://github.com/jaredpalmer/razzle/pull/1112#issuecomment-724397931

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Razzle           | ^3.3.7
| Razzle Plugins   | N/A
| Node             | 12.18.3
| Browser          | N/A
| npm/Yarn         | 1.22.5
| Operating System | Ubuntu 20.08
| TypeScript       | ^3.8.3
| React            | ^16.13.1

Edit: I found a workaround, mentioned here: https://github.com/jaredpalmer/razzle/pull/1112#issuecomment-724956179

parkerault:
For context, see @fivethreeo's comments here: https://github.com/jaredpalmer/razzle/pull/1112#issuecomment-724397931

parkerault:
Turns out this was a failure of prior knowledge on my part, and can be fixed by an update to the documentation: #1468. I'll close this issue.

Issue #1410, opened 2020-09-22
Documentation


fivethreeo:
With regards to documentation.

Specifically on the canary branch, what do you think is missing or needs more elaboration?

Any pointers or pull requests greatly apprechiated :)

alexanderankin:
does modify still work? docs should reflect that

fivethreeo:
documented the change now

https://razzle-git-canary.jared.vercel.app/docs/customization#extending-webpack

alexanderankin:
@fivethreeo can you tell where and what changes? As far as i can tell [this should still work](https://razzle-docs.vercel.app/docs/customization#extending-webpack) but it didnt for me.

fivethreeo:
do you use the finch/canary or stable release. Do a `yarn why razzle`.

alexanderankin:
@fivethreeo here's the output, i just used the yarn create functionality:

```
yarn why v1.22.5
[1/4] Why do we have the module "razzle"...?
[2/4] Initialising dependency graph...
[3/4] Finding dependency...
[4/4] Calculating file sizes...
=> Found "razzle@4.0.0-finch.8"
info Has been hoisted to "razzle"
info This module exists because it's specified in "devDependencies".
info Disk size without dependencies: "1.9MB"
info Disk size with unique dependencies: "1.9MB"
info Disk size with transitive dependencies: "1.9MB"
info Number of shared dependencies: 272
Done in 1.17s.
```

fivethreeo:
https://razzle-git-finch.jared.vercel.app/docs/customization#extending-webpack

smcyr:
There seems to be missing options for static export in the canary documentation:
https://razzle-git-canary.jared.vercel.app/docs/customization#to-enable-static-export
I can see here that there is the `paralell` options:
https://github.com/jaredpalmer/razzle/blob/canary/packages/razzle/scripts/export.js#L155
(It also has a typo, it should be `parallel` and not `paralell`.

nimaa77:
ok, I'm going to add some future todos here

- [x] add docs for razzle static
- [x] add docs Absolute Imports (#1206)
- [x] move plugins docs to docs website
- [x] add docs for deploying razzle apps on azure, amazon, and others ...

parkerault:
@nimaa77 I created a PR for the non-relative imports documentation here: #1468. Let me know if it meets your expectations.

andyzeli:
[...]

Issue #1206, opened 2020-03-10
Absolute Imports


itsemirhanengin:
Hello,

I need absolute imports. I don't want to use "../../../../components/Button" every single time. I want to use like "~/components/Button". Is it possible? I found an thing for this but it works like "components/Button" i want to add prefix.

Best,

nimaa77:
@itsemirhanengin hello, check #1112 

we are going to have a new version of razzle in this or next week

nimaa77:
oh sorry didn't want to close this :)

itsemirhanengin:
I'm still using like that but sometimes eslint can't handle that because it seems like a package. That's why i want to make prefixes actually. I'm excited about new version :)

nathanburgess:
Just started working with Razzle today and #1112 doesn't seem to work for me.

itsemirhanengin:
It should've work. Could you send me your config codes that you wrote to razzle.config.js?

nimaa77:
Hello
I merged branch next to develop

and this is what happened :)
CI/CD pipeline failed

so new version didn't deploy on npm

this is why you can't use this

nimaa77:
you can use this feature in razzle 3.1

ost12666:
I just used it, its great! The only problem is that this and other 3.1 features are not updated in the project readme...

nimaa77:
this issue fixed, closing now

(@ost12666 I will add docs for absolute imports)

PR #1468, opened 2020-11-10
Add documentation for non-relative module imports.

parkerault:
Non-relative module support isn't currently documented except for a comment in #1112. I wasn't aware that you could also set module base directories in package.json -> jest, so it took me a lot of back and forth before I finally figured it out. I think rather than trying to automate the jest configuration based on the `baseUrl` setting, an update to the documentation is a more appropriate solution. This addresses #1466.

Edit: I just found discussions in #1410 and #1206 as well.

### DIFF
--- a/website/pages/docs/customization.md
+++ b/website/pages/docs/customization.md
@@ -20,6 +20,40 @@ module.exports = {
 
 More options will be added in the future, some options now in experimental will be moved here.
 
+## Absolute Imports
+
+You can configure your application to support importing modules using absolute paths. This can be done by configuring a jsconfig.json or tsconfig.json file in the root of your project. If you're using TypeScript in your project, you will already have a tsconfig.json file.
+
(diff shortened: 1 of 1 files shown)


nimaa77:
hey, looks good to me

just a side note: let's change the title from `Non-Relative Module Imports` to `Absolute Imports`, so it's more consistent with [create-react-app docs](https://create-react-app.dev/docs/importing-a-component/#absolute-imports) (since most the people are familiar with it and it's easier for those people to map the things together)

parkerault:
Updated 👍

nimaa77:
thanks and GG

nimaa77:
@fivethreeo should we merge this into master?

fivethreeo:
We should add it to both master and canary. Add to canary and cherry pick or other way around if possible. Just make sure it gets added to both.

fivethreeo:
Also it should mention compilerOptions.paths.

fivethreeo:
Added to docs :)

PR #1476, opened 2020-11-19
Use Dart sass instead of deprecated node-sass

slobo:
Webpack sass-loader recommends this, node-sass doesn't work with PnP either.

Dart sass tends to install faster too, as it doesn't need to compile libsass as far as I understand

### DIFF
--- a/packages/razzle-plugin-scss/package.json
+++ b/packages/razzle-plugin-scss/package.json
@@ -13,12 +13,12 @@
     "autoprefixer": "10.0.1",
     "css-loader": "5.0.0",
     "deepmerge": "4.2.2",
-    "node-sass-chokidar": "1.5.0",
     "postcss-flexbugs-fixes": "4.2.1",
     "postcss-load-config": "3.0.0",
     "postcss-loader": "4.0.4",
     "postcss-scss": "3.0.2",
     "resolve-url-loader": "3.1.2",
+    "sass": "^1.29.0",
     "sass-loader": "10.0.3"
   },
   "devDependencies": {


fa7ad:
Hi @jaredpalmer ! Sorry to be that guy but any updates on when this can be merged?

fivethreeo:
This patch will not be merged. Please apply it to canary instead.

slobo:
Ok, applied it to `canary`. Let me know if there is anything else I can do. Thanks

fivethreeo:
Can you fix the commit message makes changelog better

https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit

slobo:
Does this work? Thanks

fivethreeo:
Great :)

fivethreeo:
I want to add this, but not sure if there is any backwards compatibility issues and platform support,

slobo:
In our tests, it seems to be a drop-in replacement, and even webpack `sass-loader` recommends this.
https://webpack.js.org/loaders/sass-loader/

image

johnrom:
Would be nice to have a `canary` release for `v3`, and a `next` release for `v4`. I'm afraid to see what happens when I try opting into v4!

🙏

fivethreeo:
I am testing install/build/start of all examples in v4.0.0-canary now to make sure all work. I really don’t want to touch v3 it has too little tests and outdated dependencies. So fixing stuff is hard without breaking things. I will do a @three tag so you can opt out 😀

johnrom:
I can't update to v4 because it requires eslint 6 which another dependency (prettier-eslint-cli) doesn't support. Such is life!

On the other hand, you don't have to update to get this working. You can instead do:

```
npm install node-sass@npm:sass sass
```

This will alias node-sass for anyone explicitly importing `node-sass`, while letting `sass-loader` know that it can actually use `sass` instead (and use any optimizations necessary).

Even after all that, you'll find migrating an existing project to sass modules to be even harder.

fivethreeo:
I could do a razzle-plugin-eslint6 in v4 if that helps.

fivethreeo:
Um, 3.x uses eslint 6, 4.x uses 7

johnrom:
I assume 3.x uses 6 but it works with 5, but 4.x does not work with 5. I don't have time to redo the installs and double check. All I know is I got a ton of errors when I updated, and it was easier to alias node-sass to dart sass as they have the same API. I don't need anything done on my end since the alias works fine for now, which I didn't realize was possible when I requested backporting this change into v3.

Issue #1499, opened 2020-12-31
Issues with node externals


therealgilles:
## ❓Question
I'm using the simple-oauth2 node module on the server side. It uses the (new) [private] class field syntax. Webpack tells me that it cannot parse the associated files as follows:

```
throw new Error("Module parse failed: Unexpected character '#' (19:2)\nYou may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders\n| \n| module.exports = class Client {\n>   #config = null;\n|   #client = null;\n| ");
      ^
Module parse failed: Unexpected character '#' (19:2)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
| 
| module.exports = class Client {
>   #config = null;
|   #client = null;
|
    at Object.../server/node_modules/simple-oauth2/lib/client/client.js (.../app/build/server.js:507057:7)
    at __webpack_require__ (,,,/app/webpack/bootstrap:754:1)
```

Is that something that's supposed to be supported?
If so, where am I going wrong in my config? babel/webpack?
I tried razzle@canary but no luck either.

Thanks for any insight and advice :)

fivethreeo:
In canary you can use modifyBabelPreset and add https://babeljs.io/docs/en/babel-plugin-proposal-private-methods to Babel plugins

fivethreeo:
Or simpler, add that to plugins in .babelrc

therealgilles:
@fivethreeo, thanks for the suggestion. I tried adding it to .babelrc but the result was the same.

therealgilles:
I realize now I should (1) have filed this under Discussions, and (2) explain my repo structure a bit more. So here we go:

I'm trying to add razzle to an existing create-react-app repo using redux. My repo structure looks like this:

```
my-app
|-- node_modules
|-- app
     |-- node_modules
     |-- src
     |-- index.js (client)
|-- server
     |-- node_modules
          |-- simple-oauth2
     |-- src
     |-- index.mjs (http server)
     |-- expressServer.mjs (express server)
|-- config
     |-- config.mjs (server config file)
```

I added razzle (and its dependencies) directly under my-app and then configure the path using razzle.config.js.

```JS
// razzle.config.js

const path = require('path')

module.exports = {
  options: {
    verbose: true,
  },

  modifyPaths({
    webpackObject,
    options: {
      razzleOptions,
    },
    paths,
[...]

PR #1503, opened 2021-01-03
fix(razzle): Use 'request' when externalizing modules for node externals

therealgilles:
@fivethreeo: I hope this is alright.

See https://github.com/jaredpalmer/razzle/issues/1499.


fivethreeo:
Yes all ok 😀

PR #1521, opened 2021-01-12
Fix babel-jest with babelrc

CarsonF:
`createTransformer` used to be able to accept `presets: false` but with [this](https://github.com/facebook/jest/pull/9766/files#diff-5f463bc5dbe428998302335614e9bbed3b4eb3310a6075767e1b432c161d9e71R56) change it now requires `null` or `undefined` (logic changed from `||` to `??`)


fivethreeo:
Can you change the commit message to

“fix(razzle): message”

Issue #1509, opened 2021-01-06
How to get css assets in the server rendered html


therealgilles:
## ❓Question

I've looked at a few examples (like --with-single-exposed-port) and the App.css is not been added to the server rendered html. It's only added once the client.js file is interpreted by the browser. There should be a way to have css added as link resources in the header (I see the code is there for that).

If someone could clarify how that's supposed to work, I would be grateful.

fivethreeo:
Look at the basic example and copy that. If you can fix the example can you do a pull request?

therealgilles:
There seems to be a couple solutions here:

1) Use the isomorphic-style-loader plugin as described [here](https://medium.com/@danielnmai/load-css-in-react-server-side-rendering-with-isomorphic-style-loader-848c8140a096). This requires using the plugin in the App code.

2) Use the file-loader plugin for css/sass/scss files. This requires using `import styles from './App.css'` instead of `import './App.css'`. See [here](https://dev.to/achhapolia10/adding-style-to-server-side-rendering-and-automating-the-build-process-59l).

Still researching... will try to put a separate example together.

fivethreeo:
Basic example is working, no loaders are needed :)

fivethreeo:
https://github.com/jaredpalmer/razzle/blob/canary/examples/basic/src/server.js#L8

therealgilles:
I'm looking at the basic example and I don't see any link tag stylesheet added to the generated html. My understanding is the styles are added when the client.js script is loaded.

PS: I am looking at the non-canary version right now.

fivethreeo:
https://github.com/jaredpalmer/razzle/blob/master/examples/basic/src/server.js#L21

therealgilles:
I see this code but assets.client.css is empty, so nothing is getting added.

fivethreeo:
In development css is loaded in js.

therealgilles:
Oh I see. May I ask why?

fivethreeo:
If you run the build version it is added

fivethreeo:
So hot reloading works in dev

therealgilles:
Is there a way to test it's working in development mode?

fivethreeo:
I have had some issues with that aswell, don't know really :|

therealgilles:
I think I got it working by hacking createConfigAsync.js. Let me see if I can come up with a razzleConfig.js setup just for testing.

therealgilles:
[...]

PR #1535, opened 2021-01-30
feature(razzle): new writeCssDev razzle option

therealgilles:
resolves jaredpalmer/razzle#1509

This option adds stylesheet link tags in the header for the development build, similar to the production build. It allows to
test this functionality in development (and avoid the Flash Of Unstyled Content effect), but prevents hot reloading when the css is updated. It is turned off by default.

### DIFF
--- a/packages/razzle/config/createConfigAsync.js
+++ b/packages/razzle/config/createConfigAsync.js
@@ -526,7 +526,8 @@ module.exports = (
             ]
             : IS_DEV
             ? [
-              require.resolve('style-loader'),
+              razzleOptions.writeCssDev
+                ?  MiniCssExtractPlugin.loader : require.resolve('style-loader'),
               {
                 loader: require.resolve('css-loader'),
                 options: {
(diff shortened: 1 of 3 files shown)


packages/razzle/config/createConfigAsync.js
```diff
@@ -806,6 +813,7 @@ module.exports = (
               })
             : null,
           new webpack.DefinePlugin(webpackOptions.definePluginOptions),
+          razzleOptions.writeCssDev ? miniCssExtractPlugin : null,
```
fivethreeo, line 816: razzleOptions.writeCssDev ||  IS_PROD ? miniCssExtractPlugin : null

fivethreeo:
For 5.0 I think make everything plugins.

packages/razzle/config/createConfigAsync.js
therealgilles: If I read the code correctly, I think this statement is already under IS_DEV. Do I have it wrong?

fivethreeo:
Ah, the  diff confused me. all ok

fivethreeo:
Are you on discord? If so come here https://discord.gg/pJSg287

therealgilles:
Is this a good place to move up the two options in the documentation? I also updated the comment.

```JS
    cssPrefix: 'static/css',
    jsPrefix: 'static/js',
    mediaPrefix: 'static/media',
    staticCssInDev: false, // static css in development build (incompatible with css hot reloading)
    browserslist: undefined, // or what your apps package.json says
    enableReactRefresh: false,
    enableTargetBabelrc: false, // enable to use .babelrc.node and .babelrc.web
```

fivethreeo:
Yes

PR #1559, opened 2021-02-22
feat(razzle): disable prompt on CI

SimeonC:
If the CI env var is set we should not show the prompt as the CI is by default not interactive.


SimeonC:
Fixes conversation https://github.com/jaredpalmer/razzle/commit/d86aaca2f3ec444c633219513dd0ec4f26d65614

PR #1560, opened 2021-02-22
fix(razzle-dev-utils): correct webpack 4 format

### LINKED ISSUE #1558: razzle-dev-tools v4 Issues with error/warning output with Webpack 4
With the canary.47 changes I found that my client bundle build warnings and errors are not being printed to the console.

From the looks of it, it's because in webpack 4 the warnings are just strings, not objects so the `printWarnings` function basically prints nothing.

I tested by adding the following code to the top of the printWarnings function and it output the information I needed.
```
if (!wrn.message && !wrn.details && !verbose) {
  console.warn(wrn);
}
```
https://github.com/jaredpalmer/razzle/blob/a1b4def5f3b1af66bc0d6f472f7c497606f7eb1e/packages/razzle-dev-utils/printWarnings.js#L14

As for the printErrors function, I tested by running `CI=true razzle build` and it looks like the `err` value is an array of strings 🤷 
The suitable catch for this I think is just checking for `err.length`, that seemed to work when testing locally.
```
if (err.length) {
  console.error(err.join(''));
}
```
https://github.com/jaredpalmer/razzle/blob/a1b4def5f3b1af66bc0d6f472f7c497606f7eb1e/packages/razzle-dev-utils/printErrors.js#L15

SimeonC:
Webpack 4 has a different output format for warnings and errors than webpack 5. Weirdly enough this is a string and array of strings.

Closes #1558

### DIFF
--- a/packages/razzle-dev-utils/printErrors.js
+++ b/packages/razzle-dev-utils/printErrors.js
@@ -12,12 +12,17 @@ function printErrors(summary, errors, verbose) {
   console.log(chalk.red(summary));
   console.log();
   errors.forEach(err => {
-    if (err.message) {
-      console.error(err.message);
-    }
-    console.error(err.stack || err);
-    if (err.details) {
-      console.error(err.details);
+    if (Array.isArray(err)) {
+      // webpack 4 format
+      console.error(err.join(''));
+    } else {
+      if (err.message) {
+        console.error(err.message);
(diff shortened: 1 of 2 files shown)


fivethreeo:
Broke in webpack 5, but a good catch will fix it up a bit and it will work :)

PR #1597, opened 2021-04-15
examples: add esbuild-loader example

rclmenezes:
We started using esbuild-loader instead of ts-loader and it's been amazing.

This example:
- Copied from `with-typescript-plugin`
- Replace `razzle.config.js` with contents of packages/razzle-plugin-typescript, with some modifications
- Replace ts-loader with esbuild-loader

Why esbuild-loader instead of babel?
1. Performance. We saw a 2x improvement over ts-loader on `yarn razzle start`.
2. Babel still does some weird things to Typescript. My personal favorite [was this gem](https://github.com/Vincit/objection.js/issues/1791#issuecomment-815407687), where babel decided to tinker with a constructor and make some properties defined with '!:` become undefined... after the `super()`! Oof, what a mess.

Also, I added a commit to fix a typo where some examples said "Welcome to Razzles" lol


rclmenezes:
image

PR #1617, opened 2021-05-05
Make package.json optional

squiddy:
In some setups there may only be one package.json in the root directory,
and multiple apps (using RAZZLE_APP_PATH) are developed in separate
subdirectories.

Before this change, razzle expected a package.json to be present in each
directory - even if it contained an empty object.

---

See the discussion in https://github.com/jaredpalmer/razzle/discussions/1611. Marking it as draft because I want to look into tests still, might need some help there.

The check/clearConsole/logging part I grabbed from `packages/razzle/config/loadRazzleConfig.js` which was already doing that.


squiddy:
I'm wondering what is more appropriate, a test case or an example (looks like those are also tested)?

fivethreeo:
A example would do nice :)

squiddy:
> A example would do nice :)

I added an example. Some thoughts:

* I can't come up with a good name right now
* README says to run `yarn update-examples`, but that changes all the READMEs for me. Probably messed up my local setup?

fivethreeo:
Should update all readmes, but it checks the branch name. So only run it on releases.

fivethreeo:
I merged it but regular tests fail. Not sure why yet ;)

squiddy:
> I merged it but regular tests fail. Not sure why yet ;)

Sorry about that. I'm afraid I won't have time today to look into this, I can have a look tomorrow.

squiddy:
I'll have a PR with a fix soon.

fivethreeo:
https://github.com/jaredpalmer/razzle/blob/bae26bfd509570bb05f3059bcc220ed86434d00e/packages/razzle/config/paths.js#L34 I guess 😀

squiddy:
> https://github.com/jaredpalmer/razzle/blob/bae26bfd509570bb05f3059bcc220ed86434d00e/packages/razzle/config/paths.js#L34
> I guess grinning

Exactly, not sure yet how I managed to mess that up. Adding a `const ` and running everything test-related locally looks fine.

Bit tied up at work right now, still happy to provide a fix if you'd like.

PR #1620, opened 2021-05-07
fix(razzle): Report config errors for start command

squiddy:
Currently, if code inside the razzle.config.js (e.g.modifyWebpackConfig) 
throws an error, the process just quits without displaying any error message.

Long-term it might be better to catch this error a bit closer to the
code that is causing it (e.g. "modifyWebpackConfig raised an error"),
but for now this aligns the handling to what the build and test command
are already doing.

### DIFF
--- a/packages/razzle/scripts/build.js
+++ b/packages/razzle/scripts/build.js
@@ -1,13 +1,6 @@
 #! /usr/bin/env node
 'use strict';
 
-// Makes the script crash on unhandled rejections instead of silently
-// ignoring them. In the future, promise rejections that are not handled will
-// terminate the Node.js process with a non-zero exit code.
-process.on('unhandledRejection', err => {
-  throw err;
-});
-
 const mri = require('mri');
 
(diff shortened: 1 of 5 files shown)


fivethreeo:
The test added fails on windows with wrong returncode.

When tests fail it is usually on windows 🎻

squiddy:
🤦 Windows.. okay, i'll have a look. :)

squiddy:
:man_shrugging: I ran the workflows multiple times in my fork and I can see windows tests randomly failing, but my test not once.

fivethreeo:
Strange, fails every time for me on github CI. May be some caching.. Probably not a issue. I did not try win tests locally.

squiddy:
> May be some caching.. Probably not a issue.

Looks like it failed again on master, I'll try to setup a windows machine.

fivethreeo:
Passes now after yarn.lock was updated. Was cache somehow..

PR #1622, opened 2021-05-07
ci: Add github workflow for running prettier

squiddy:
Prettier is configured for the repository but not all code is actually
formatted. This workflow runs also for pull requests and can ensure code
stays formatted.

---

I'm not actually familiar with github workflows, so I might be doing it not entirely correct. It does work fine on my fork though. I wasn't sure initially whether the project wants something like that, but my editor kept formatting code I didn't touch, so maybe this helps.

*Update* Oh, it even runs for this pull request now. :astonished:

### DIFF
--- a/.github/workflows/prettier.yml
+++ b/.github/workflows/prettier.yml
@@ -1,0 +1,23 @@
+name: Prettier
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+  workflow_dispatch:
+
+jobs:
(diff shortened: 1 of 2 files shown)


fivethreeo:
Not sure about this one, need to think about it 😀

squiddy:
> Not sure about this one, need to think about it grinning

Sure, just curious: is this about prettier in general, the workflow or running CI for pull requests in general (wondering especially about this one since I tend to break stuff which is a bit frustrating)?

PR #1654, opened 2021-07-06
Fix mini css extract plugin with webpack5

wokayme:
## Overview problem

During the installation of Webpack 5 I got an error:
Screenshot 2021-07-06 at 10 05 44

Problem is that version of `mini-css-extract-plugin: 0.9.0 has in peerDependencies` exact version
```
"peerDependencies": {
    "webpack": "^4.4.0"
  },
 ```
 
 So why we won't allow users to bump it to a higher version like `0.12.0`
 ```
 "peerDependencies": {
    "webpack": "^4.4.0 || ^5.0.0"
  },
  ```
  Which gives us support of webpack 5?

  
  `^0.9.0` won't allow us to bump to a higher version (https://semver.npmjs.com/)
Screenshot 2021-07-06 at 12 25 49

  
   so why don't allow us to bump to `0.12.0` by updating the version condition?
   
Screenshot 2021-07-06 at 12 26 02

### DIFF
--- a/packages/razzle-plugin-less/package.json
+++ b/packages/razzle-plugin-less/package.json
@@ -25,7 +25,7 @@
   },
   "peerDependencies": {
     "less": "^4.1.0",
-    "mini-css-extract-plugin": "^0.9.0",
+    "mini-css-extract-plugin": ">=0.9.0 <1.0.0",
     "postcss": "^8.2.4",
     "razzle": "4.0.5",
     "razzle-dev-utils": "4.0.5",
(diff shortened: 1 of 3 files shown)


fivethreeo:
Need to update razzle.plugin-scss/less aswell

can you make the commit message: fix(deps): ....

https://github.com/angular/angular/blob/master/CONTRIBUTING.md#-commit-message-guidelines

wokayme:
Name fixed, and bumped also in plugins, I verified changes in examples and everything looks working

PR #1735, opened 2021-10-20
Follow-up fix for #1704 to remove some warnings

heath-freenome:
- Updated `packages/razzle/package.json` to allow `webpack-dev-server` v4 as a peerDependency
- Updated `createConfigAsync.js` to move the `hot: true` flag into just the v3 config since it is on by default for v4 removing the following warning:
```
[webpack-dev-server] "hot: true" automatically applies HMR plugin, you don't have to add it manually to your webpack configuration.
```
- Also fixed a small typo in a comment

### DIFF
--- a/packages/razzle/config/createConfigAsync.js
+++ b/packages/razzle/config/createConfigAsync.js
@@ -851,7 +851,6 @@ module.exports = (
             disableDotRule: true,
           },
           host: dotenv.raw.HOST,
-          hot: true,
           port: devServerPort,
         };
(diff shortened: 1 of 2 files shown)


heath-freenome:
@fivethreeo here is that PR I mentioned

heath-freenome:
@fivethreeo I would hold off releasing this as I'm still trying to solve a few of the other warnings... including the fact that moving the `hot: true` ends up being the wrong fix

PR #1743, opened 2021-10-25
Simple plugin to load markdown files transpiled to html

uebelack:
Added a simple plugin that uses markdown-loader to load markdown files as html via webpack.


fivethreeo:
There already is a mdx plugin. But the example is good, could you update the with mdx example instead?

uebelack:
Yeah, didn't know MDX ;-) have adjusted the PR and as suggested only extended the docu and example for MDX.

PR #1744, opened 2021-10-25
Allow create in empty folder

RobertSandiford:
How's this?

### DIFF
--- a/packages/create-razzle-app/lib/index.js
+++ b/packages/create-razzle-app/lib/index.js
@@ -31,8 +31,20 @@ module.exports = async function createRazzleApp(opts) {
   }
 
   if (fs.existsSync(projectName)) {
-    console.log(messages.alreadyExists(projectName));
-    process.exit(1);
+    const dir =  fs.opendirSync(projectName)
+    const firstFile = dir.readSync(); const secondFile = dir.readSync()
+    dir.close()
+    if ( ! ( // require the folder to be empty or have only a .git folder
+      firstFile === null 
+      || ( 
+        firstFile.name === '.git'
+        && fs.lstatSync(projectName+'/'+firstFile.name).isDirectory() 
+        && secondFile === null
+      ) 
+    ) ) {
+      console.log(messages.folderNotEmpty(projectName));
+      process.exit(1);
+    }
   }
 
   const projectPath = (opts.projectPath = process.cwd() + '/' + projectName);
--- a/packages/create-razzle-app/lib/index.js
+++ b/packages/create-razzle-app/lib/index.js
@@ -30,24 +30,12 @@ module.exports = async function createRazzleApp(opts) {
     process.exit(1);
   }
 
-  if (fs.existsSync(projectName)) {
-    const dir =  fs.opendirSync(projectName)
-    const firstFile = dir.readSync(); const secondFile = dir.readSync()
-    dir.close()
-    if ( ! ( // require the folder to be empty or have only a .git folder
-      firstFile === null 
-      || ( 
-        firstFile.name === '.git'
-        && fs.lstatSync(projectName+'/'+firstFile.name).isDirectory() 
-        && secondFile === null
-      ) 
-    ) ) {
-      console.log(messages.folderNotEmpty(projectName));
-      process.exit(1);
-    }
-  }
-
   const projectPath = (opts.projectPath = process.cwd() + '/' + projectName);
+  
+  if ( fs.existsSync(projectName) && ! isSafeToCreateProjectIn(projectPath, projectName)) {
+    console.log(messages.folderNotEmpty(projectName));
+    process.exit(1);
+  }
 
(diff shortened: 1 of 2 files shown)


packages/create-razzle-app/lib/index.js
```diff
@@ -31,8 +31,20 @@ module.exports = async function createRazzleApp(opts) {
   }
 
   if (fs.existsSync(projectName)) {
-    console.log(messages.alreadyExists(projectName));
-    process.exit(1);
+    const dir =  fs.opendirSync(projectName)
+    const firstFile = dir.readSync(); const secondFile = dir.readSync()
```
fivethreeo: Not sure about depending on order of dirs.

RobertSandiford: What do you mean?

It opens the target dir if it exists. firstFile will be a file or null. secondFile will be a file or null (if the first is null, the second is also null).

The guard activates unless firstFile is null, or firstFile is a .git folder, and secondFile is null.

I'm not follow on an order of dirs?

RobertSandiford: P.S. I tested on windows. Would prefer a test on linux too. (would take me forever to set up)

fivethreeo: What if the first is .asomething second is .bsomething and third is .git?

fivethreeo: Just use path.join and test directy?

fivethreeo: https://github.com/jaredpalmer/razzle/blob/master/test/e2e/create-razzle-app.test.js

RobertSandiford: It must fail the check and stop the creation if there is more than 1 file/folder inside.

It allows strictly an empty folder, or a .git folder only (so that a user can do `git init` `npx create-razzle-app .`)
So it simply fails in that situation because secondFile is not null.

RobertSandiford: From facebook/create-react-app

Perhaps it would be better to match this behaviour

```
// If project only contains files generated by GH, it’s safe.
// Also, if project contains remnant error logs from a previous
// installation, lets remove them now.
// We also special case IJ-based products .idea because it integrates with CRA:
// https://github.com/facebook/create-react-app/pull/368#issuecomment-243446094
function isSafeToCreateProjectIn(root, name) {
  const validFiles = [
    '.DS_Store',
    '.git',
    '.gitattributes',
    '.gitignore',
    '.gitlab-ci.yml',
    '.hg',
    '.hgcheck',
    '.hgignore',
    '.idea',
    '.npmignore',
    '.travis.yml',
    'docs',
    'LICENSE',
    'README.md',
    'mkdocs.yml',
    'Thumbs.db',
  ];
  // These files should be allowed to remain on a failed install, but then
  // silently removed during the next create.
  const errorLogFilePatterns = [
    'npm-debug.log',
    'yarn-error.log',
    'yarn-debug.log',
  ];
  const isErrorLog = file => {
    return errorLogFilePatterns.some(pattern => file.startsWith(pattern));
  };

  const conflicts = fs
    .readdirSync(root)
    .filter(file => !validFiles.includes(file))
    // IntelliJ IDEA creates module files before CRA is launched
    .filter(file => !/\.iml$/.test(file))
    // Don't treat log files from previous installation as conflicts
    .filter(file => !isErrorLog(file));

  if (conflicts.length > 0) {
    console.log(
      `The directory ${chalk.green(name)} contains files that could conflict:`
    );
    console.log();
    for (const file of conflicts) {
      try {
        const stats = fs.lstatSync(path.join(root, file));
        if (stats.isDirectory()) {
          console.log(`  ${chalk.blue(`${file}/`)}`);
        } else {
          console.log(`  ${file}`);
        }
      } catch (e) {
        console.log(`  ${file}`);
      }
    }
    console.log();
    console.log(
      'Either try using a new directory name, or remove the files listed above.'
    );

    return false;
  }

  // Remove any log files from a previous installation.
  fs.readdirSync(root).forEach(file => {
    if (isErrorLog(file)) {
      fs.removeSync(path.join(root, file));
    }
  });
  return true;
}
```

fivethreeo: Yeah, that seems good 😀

RobertSandiford:
@fivethreeo 

Facebook's logic implemented as discussed here: https://github.com/jaredpalmer/razzle/pull/1744/files/07d8169b4f91126c61c55289ee6f2a370cb9052a#r735844589

My local tests work when creating in a non-existing dir, and existing dir, and in '.'

The messages are a bit meh for '.':
```
npx create-razzle-app .

Creating ....

> Success! Created files for "." razzle app
```
But whatever

Please review

fivethreeo (approved):

RobertSandiford:
Cool. Hope it works : )

PR #1754, opened 2021-11-09
Do not wrap build errors in array before printing

### LINKED ISSUE #1753: Errors for `yarn run build` are not displayed correctly
### Current Behavior

When running `npm run build` and build errors, for example syntax errors, are found, these are displayed as an array.

```
Failed to compile client default build.

[
  './src/App.js\n' +
    'Module build failed (from ./node_modules/razzle/config/babel-loader/razzle-babel-loader.js):\n' +
    'SyntaxError: /Users/hakan/code/raderas/my-app/src/App.js: Missing semicolon. (13:4)\n' +
    '\n' +
    '\x1B[0m \x1B[90m 11 |\x1B[39m\x1B[0m\n' +
<<<  text deleted >>>
    '    at Object.parse (/Users/hakan/code/raderas/my-app/node_modules/@babel/parser/lib/index.js:10246:18)\n' +
    '    at parse (/Users/hakan/code/raderas/my-app/node_modules/@babel/parser/lib/index.js:14984:26)\n' +
    ' @ ./src/client.js 2:0-24 6:41-44'
]
```

### Expected behavior

Expected array to be unwrapped and printed as a string:

```
Failed to compile client default build.

./src/App.js
Module build failed (from ./node_modules/razzle/config/babel-loader/razzle-babel-loader.js):
SyntaxError: /Users/hakan/code/raderas/my-app/src/App.js: Missing semicolon. (13:4)

  11 |
  12 | export default App;
> 13 | just some text
     |     ^
<<<  text deleted >>> 
```

### Reproducible example

Just create an app, add some random characters to the end of src/App.js and build:

```sh
npx create-razzle-app my-app
cd my-app
echo 'just some text' >> src/App.js
npm run build -- --noninteractive
```

### Cause

The problem is in https://github.com/jaredpalmer/razzle/blob/v4.2.6/packages/razzle/scripts/build.js#L112-L114 where the array of build errors is used to reject the promise.

``` js
if (clientMessages.errors.length) {
  return reject(clientMessages.errors);
}
```

This "error" is catched  in https://github.com/jaredpalmer/razzle/blob/v4.2.6/packages/razzle/scripts/build.js#L186-L193 and wrapped in another array. 

``` js
err => {
  printErrors(
    `Failed to compile client ${buildName} build.`,
    [err],
    verbose
  );
  rejectBuild(buildName, 'web');
}
```

`printErrors` https://github.com/jaredpalmer/razzle/blob/v4.2.6/packages/razzle-dev-utils/printErrors.js loops over the elements and for each element logs it (in this case it logs the array)

### Suggested solution(s)
The best would be to rewrite https://github.com/jaredpalmer/razzle/blob/v4.2.6/packages/razzle/scripts/build.js so that it resolves with errors the same way it resolves with warnings.
So:
``` diff
  if (clientMessages.errors.length) {
-   return reject(clientMessages.errors);
+   return resolve({ errors} );
  }
```
And in https://github.com/jaredpalmer/razzle/blob/v4.2.6/packages/razzle/scripts/build.js#L153 :

```diff
- ({ stats, previousFileSizes, warnings }) => {
+ ({ stats, previousFileSizes, warnings, errors }) => {
+   if(errors && errors.length) {
+     printErrors(
+       `Failed to compile client ${buildName} build.`,
+       errors,
+       verbose
+     );
+     return rejectBuild(buildName, 'web');
+   }
    if (warnings.length) {
```

The same should be don in the code handling server builds as well in https://github.com/jaredpalmer/razzle/blob/v4.2.6/packages/razzle/scripts/build.js

Another option is to flatten the array in https://github.com/jaredpalmer/razzle/blob/v4.2.6/packages/razzle-dev-utils/printErrors.js#L15 (hacky)
```diff
- errors.forEach(err => {
+ errors.flat().forEach(err => {
```

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Razzle           | 4.2.6
| Node             | v14.16.0
| Operating System | OSX 11.6

HCanber:
This applies to the `build.js` script that's executed on `yarn build`

## How it currently works
When build errors occurs (or warnings for CI builds) the build promise is rejected with an array – the error messages, see https://github.com/jaredpalmer/razzle/blob/v4.2.6/packages/razzle/scripts/build.js#L112-L114 for an example.

When handling the rejected promise, the error is wrapped in an array before calling `printErrors` as it's expects an array of errors to print. This is fine when the promise is rejected with an actual `Error` but for build errors this means that the array gets wrapped in a new array, and instead of getting an array of _n_  messages, `printErrors` receives an array with 1 element and it will log the array instead of the indivdual messages, see #1753 for an example on what this looks like.

## This PR
There is no need to wrap the array of error messages in a new array before calling printErrors, so in the catch clause if   `err` is an array, it will not be wrapped. If it's not an array, it's wrapped, just like previously.

## Verification
I've manually verified this PR with webpack 4 for `examples/basic` and webpack 5 for `examples/with-preact`. Syntax errors where introduced to the examples with this:

``` sh
echo 'just some text' >> examples/basic/src/App.js
echo 'just some text' >> examples/with-preact/src/App.js
```

And then `yarn build --noninteractive` was ran for each example.

Of course `yarn test --runInBand` was ran as well.

Fixes #1753

### DIFF
--- a/packages/razzle/scripts/build.js
+++ b/packages/razzle/scripts/build.js
@@ -186,7 +186,7 @@ loadRazzleConfig(webpack).then(
                     err => {
                       printErrors(
                         `Failed to compile client ${buildName} build.`,
-                        [err],
+                        Array.isArray(err) ? err : [err],
                         verbose
                       );
                       rejectBuild(buildName, 'web');
@@ -285,7 +285,7 @@ loadRazzleConfig(webpack).then(
                   err => {
                     printErrors(
                       `Failed to compile server ${buildName} build.`,
-                      [err],
+                      Array.isArray(err) ? err : [err],
                       verbose
                     );
                     rejectBuild(buildName, 'node');


fivethreeo:
I will pull this but can you change the commit msg

`git reset —soft HEAD^`
`git commit -a -m "fix(razzle): make errors log correctly in build"`
`git push -f`

Issue #1159, opened 2019-11-13
Looking for help / maintainers (Please read)


elliottjro:
It seems like there has been very low traction on this project @jaredpalmer 

cc @maintainers-of-razzle (woops, pinged `maintainers` by accident)

There are many PRs with no comments from any of the maintainers, is there anything we can do to get this project back up and running?

EDIT:
I will start working on the PRs from oldest. If there is anything that is super urgent please just @ me to that PR / Issue and I will try to get to it in a timely manner

fifn2:
Uh oh...

fivethreeo:
No answer yet. Anyone up for forking to “dazzle” if another week goes by?

elliottjro:
@fivethreeo that is the plan

ivywitter:
@elliottjro any update on this?

jaredpalmer:
I’m not actively working on Razzle at the moment. It’s not that I’ve moved onto other alternatives, it’s just that I’m not currently working on any SSR applications. It is thus a lower-priority project for me until 1) React’s new server renderer comes out and 2) webpack 5 drops. If folks want to step up and help out that would be rad. If you’re interested, send me a DM on Twitter.

fifn2:
React is getting a new server renderer? Where can I find out more about this?

jgcmarins:
@fifn2 here: https://twitter.com/jgcmarins/status/1200172819994357761

elliottjro:
@jaredpalmer dmed you on twitter

licg9999:
How is it going? Webpack 5 beta released.

elliottjro:
well @jaredpalmer is not responding on twitter or on github sooo we are just stuck here in limbo

jaredpalmer:
Just saw your message Elliot. Adding you as a contributor

licg9999:
Razzle is good, SSR in mind is good, wish this project can go on. One day I need to mix gatsby and  CRA, I found no better choice than Razzle.

elliottjro:
Alrighty folks, Im gonna start working on the PRs from oldest. If there is anything that is super urgent please just @ me to that PR / Issue

revskill10:
I don't use SSR to run in production. I use SSR with Razzle to produce static pages :).
Razzle is no magic to me, it's good so far. I'm happy to hear the project is going on.

DomHoney:
[...]

PR #1830, opened 2022-02-22
Fix default host

romanlex:
Default host is 0.0.0.0 by doc
image


fivethreeo:
I think update docs instead of the code.

romanlex:
@fivethreeo ok, done

gregmartyn:
`0.0.0.0` is not `localhost`. `Localhost` is `127.0.0.1`; `0.0.0.0` is `INADDR_ANY`. The difference is that `0.0.0.0` listens on all network adapters, but `127.0.0.1` only listens for machine-local connections.

If you're running your site in a container/vm, your host will only be able to connect to the container if the container is listening on `0.0.0.0` ~and a port forward has been set up~. If the container is listening on `127.0.0.1`, it won't be able to connect -- only connections from the container itself can connect to it.

If you're not running inside a VM, using `127.0.0.1` is a little like having a firewall: other computers on your network won't be able to connect to it no matter what. `0.0.0.0` would allow those connections if you don't have a firewall configured.

More info: https://superuser.com/questions/949428/whats-the-difference-between-127-0-0-1-and-0-0-0-0

romanlex:
@gregmartyn it's pr changes only md files for consistency

gregmartyn:
I don't see how it's making it consistent though. Razzle listens on `0.0.0.0` by default as far as I can tell.

https://github.com/jaredpalmer/razzle/blob/master/packages/create-razzle-app/templates/default/src/index.js
does
```
express()
    .listen(port, function(err) {
```
which is https://expressjs.com/en/api.html#app.listen
which is https://nodejs.org/api/http.html#http_server_listen
which is https://nodejs.org/api/net.html#serverlistenport-host-backlog-callback
which says:
If host is omitted, the server will accept connections on the [unspecified IPv6 address](https://en.wikipedia.org/wiki/IPv6_address#Unspecified_address) (::) when IPv6 is available, or the [unspecified IPv4 address](https://en.wikipedia.org/wiki/0.0.0.0) (0.0.0.0) otherwise.

In most operating systems, listening to the [unspecified IPv6 address](https://en.wikipedia.org/wiki/IPv6_address#Unspecified_address) (::) may cause the net.Server to also listen on the [unspecified IPv4 address](https://en.wikipedia.org/wiki/0.0.0.0) (0.0.0.0).

gregmartyn:
Oh I see what you mean. Sorry!

devServer uses `HOST` here https://github.com/jaredpalmer/razzle/blob/master/packages/razzle/config/createConfigAsync.js#L854
for the port 3001 listener (aka `PORT_DEV`)
and on line 138 when `CLIENT_PUBLIC_PATH` wasn't set.

So we have a problem. `HOST` isn't being used by the prod server, but probably should be. A default of `0.0.0.0` would make sense for that, especially for backward compatibility. Otherwise you can't choose which network interface the production server listens on, which could be a problem if you want to have different services listening on different interfaces but the same port.

The dev server is defaulting to `127.0.0.1` for the `PORT_DEV` client-side build but the docs say `0.0.0.0`. (The server-side build on `PORT` behaves the same as the prod build above) `127.0.0.1` makes sense as a safe default, but then it'd conflict with the new prod server default of `0.0.0.0`. Further, the `CLIENT_PUBLIC_PATH` fallback is treating `HOST` as a routable address, but that wouldn't work for `HOST=0.0.0.0`.

Sounds to me like a separate var is in order. `HOST_DEV` (consistent with `PORT_DEV`, which already exists), defaulting to `127.0.0.1`. What do y'all think?

fivethreeo:
As long as it describes the behaviour it is all ok.

btw. I am close to stopping razzle development unless someone shares in the development. Considering just using next or plain webpack.

romanlex:
Next.js dictates how to organize a project ( routes example

sergeysova:
Hi!
How can I help you maintaining this project?
I think razzle is awesome, and I choose it again and again over Remix or Nextjs.

Thank you, for your work!

Original: https://github.com/jaredpalmer/razzle/issues/1159#issuecomment-1180418485

PR #1841, opened 2022-03-21
fixed start prod for fastify server

mindyourlifeguide:


mindyourlifeguide:
This solved problem with` '"root" option must be an absolute path'` when open localhost.
image

PR #1848, opened 2022-04-25
add support for type:module razzle.config.js

gregmartyn:
Uses `package.json`'s `type` field to determine whether to `require()` or `await import()`

The `package.json` detection is modelled on `getPublicUrl` in `razzle/packages/razzle/config/paths.js` and doesn't implement the full spec described by https://nodejs.org/api/packages.html#type ("The nearest parent package.json is defined as the first package.json found when searching in the current folder, that folder’s parent, and so on up until a node_modules folder or the volume root is reached.")

### DIFF
--- a/packages/razzle/config/loadRazzleConfig.js
+++ b/packages/razzle/config/loadRazzleConfig.js
@@ -8,10 +8,11 @@ const defaultRazzleOptions = require('./defaultOptions');
 const setupEnvironment = require('./env').setupEnvironment;
 const loadPlugins = require('./loadPlugins');
 
-const getTypeModule = appPackageJson => {
+const getModuleFormat = appPackageJson => {
   if (fs.existsSync(appPackageJson)) {
     try {
       const packageJson = require(appPackageJson);
+      // See https://nodejs.org/api/packages.html#type for more info on "type"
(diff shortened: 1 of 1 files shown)


fivethreeo:
Will not add to razzle 3.x unless it is backwards compatible. Is await in node 12+ ?

gregmartyn:
@fivethreeo `await` is already being used in that file (e.g. `razzleOptions = await plugin.modifyOptions({`) and was added in nodejs v8.

This change is backwards compatible.

In the latest commit, I noticed that that function was already requiring package.json, so this PR no longer tries to do it again.

gregmartyn:
If someone wants to opt-in to ECMAScript modules, they have to set `type: "module"` in `package.json` and convert `razzle.config.js` from commonjs to ECMAScript module. Had they tried using `type: "module"` before this PR, it would've errored because razzle was trying to `require` an ECMAScript module, but ECMAScript modules can only be `import`'ed.

There's further work to do, (e.g. setting [webpackConfig.experiments.outputModule = true](https://webpack.js.org/configuration/experiments/#experimentsoutputmodule)) so this PR alone doesn't enable ECMAScript modules, but is a step in that direction.

PR #1875, opened 2022-06-11
fix(razzle) Add "dot" option to CopyPlugin to copy hidden files

### LINKED ISSUE #1874: Razzle 4 will not copy hidden files from public dir to build dir
### Current Behavior

Inside the `public` directory, we have a hidden folder that needs to be copied into the `build/public` directory. 
FTR the hidden folder `.well-known` is for [Google's Digital Asset Links protocol,](https://developers.google.com/digital-asset-links/v1/getting-started) which needs to serve the following file: www.domain.com/.well-known/assetlinks.json 

Using Razzle v3, this hidden folder was being copied to the `build/public` directory.
Upon upgrading to Razzle v4, this folder is omitted.

### Expected behavior

Hidden folder inside `public` should also be copied to `build/public` during build.

### Suggested solution(s)

Razzle 4 makes use of `webpack-copy-plugin` to copy these files. 
There is a configuration missing in the patterns globOptions, required to also include hidden files:
`globOptions: { dot: true }`

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Razzle           | 4.2.17
| Razzle Plugins   | 
| Node             |  15.14.0
| Browser          |
| Yarn         | 1.22.5
| Operating System | MacOS
| TypeScript       | 3.8.3
| React            |

eltuza:
Adds dot config option to copy-webpack-plugin so that hidden files are also copied from public folder to the build folder.

This was the behaviour already in Razzle 3.

Closes https://github.com/jaredpalmer/razzle/issues/1874

### DIFF
--- a/packages/razzle/config/createConfigAsync.js
+++ b/packages/razzle/config/createConfigAsync.js
@@ -962,6 +962,7 @@ module.exports = (
                 context: paths.appPath,
                 globOptions: {
                   ignore: [paths.appPublic.replace(/\\/g, '/') + "/index.html"],
+                  dot: true,
                 }
               },
             ]


slobo:
:+1: We would find copying of hidden files by default useful too.

PR #1910, opened 2023-03-03
prevent calling stats.toJson()

pikpok:
Use hasErrors/hasWarnings methods to check for errors on webpack stats, instead of always calling toJson(). Also, in case of errors use `errors-only` or `errors-warnings` stat config presets to avoid generating unnecessary stats.

In our (quite large) Razzle project, `stats.toJson()` call takes a lot of time and memory, and after migrating to webpack 5 even results in OOM errors on build.


pikpok:
@fivethreeo could you take a look? 👀 or is Razzle abandoned and there's no way of getting this merged and released? 😅

rahulranjan-blitz:
looks good tbh.. need @fivethreeo approval

